PR #531, opened 2014-03-05
[NOT FUNCTIONAL] Color Button

RlonRyan:
Cycles through the colors forwards/backwards based on left click/right click.

Code is likely to break something as it was written in-browser.

Also, I have no clue how to determine left or right click... so I just guessed...
Likely will cause the conduit to cycle through the colors rapidly for no reason... if it compiles at all...

In truth, I would have preferred UML... but here is yet another failed attempt at java...

Thought I would do a little more research this time before making a request. Ten dollars says I missed something.


RlonRyan:
Ha! I knew this feature would eventually make it. Good job, I say, Good Job!

PR #768, opened 2014-06-22
Addding Tinkers Construct recipes for Alloy Smelter

hron84:
These recipes are directly from Tinkers Construct. They are technically
done, but maybe needs some more balancing in energy consumption.

### DIFF
--- a/resources/assets/enderio/config/AlloySmelterRecipes_User.xml
+++ b/resources/assets/enderio/config/AlloySmelterRecipes_User.xml
@@ -37,7 +37,7 @@ ore dictionary.
             </output>
         </recipe>
 
-        <recipe name="Alumite" energyCost="1000">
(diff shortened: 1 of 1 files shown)


MrKunji:
Endstone is not balanced, in Tinkers' is one Ender Pearl plus five Obsidian for five Endstone, or one Ender Pearl plus one Sandstone for one Endstone.

HenryLoenwind:
The obsidian recipe is has its amounts reversed. Easy to fix.

hron84:
@MrKunji I think it is changed then since I committed that change. I will take look again when I have a chance. I would like to ask a bit patience, because my laptop is currently being repaired.

tterrag1098:
As per #2418 I am changing my mind on this

PR #1432, opened 2014-12-17
Forge rotation API support

planetguy32:
Makes blocks update their heading when their rotate function is called (for example, when rotated by Remain in Motion's carriage rotator).


tterrag1098:
Reopen this if you ever fix the issues

PR #1545, opened 2014-12-31
Fixes

hilburn:
Changes powerInMachines to use long as apparently that can be a thing - prevents overflow
Fixes some typos
Changes the Set to List in config for possibly the smallest performance gain ever


tterrag1098:
Fairly straightforward. Thanks for saving me 5 minutes :P

Issue #1561, opened 2015-01-02
Farming station not accepting wooden axe.


kokono2:
The farming station doesn't accept wooden axes anymore since I updated to 1.7.10-2.2.6.317. It does accept wooden hoes and every axe better than wood.

tterrag1098:
If you have a mod which modifies the vanilla tools so that they can't mine
trees it won't work.

kokono2:
I haven't added new mods since I updated to the version of EnderIO in previous post. Mining wood with the wooden axe works perfectly fine.

Issue #1560, opened 2015-01-02
Killer Joe's and Vacuum Chests are working through solid blocks. 


Ayrrow:
Pretty self explanatory, I'm not sure if this is intended behaviour or not. It's quite annoying as I have items being pulled across rooms and getting stuck on doors or the middle of floors. 

My Killer Joe's are also damaging me through the floors and walls - this in particular seems unintended as the ability to grief other people using this is high.

hilburn:
Both are code behaving as expected - whether they should behave this way is a different matter.

Personally I would say yes it should for the vacuum chest as  every other item hoover system behaves the same way, I've had my RotaryCraft Item Vacuum pick up things from 3 chunks over...

As for the Killer Joe, maybe a config option? I think by default it would be better to leave as is - because correcting it would be relatively computationally expensive, but it would be possible to do a raytrace from the Joe to the Target and check the line of sight is clear...

tterrag1098:
Fixed in https://github.com/hilburn/EnderIO/commit/0286245a5e81316cc0de997c25fb606703d6e7ed

PR #1563, opened 2015-01-03
Bunch of fixes

hilburn:
for #1561, Wood is tool level 0, non-axes are -1, changed match function to reflect that
for  #1560, added line of sight with config to killer joe
added a null check for a crash I got intermittently when placing regular blocks next to a vibrant capacitor


Ayrrow:
Thanks for the line of sight req. I'll assume the vacuum chests are working as intended :)

hilburn:
Yeah - if I added that to the vacuum chest it would get kinda laggy - an average of about 17x more operations performed per item per tick

PR #1588, opened 2015-01-05
Added Wither Skeleton support to Powered Spawner

hilburn:
With 10*power penalty over normal - might be worth adding a config option for this...

### DIFF
--- a/src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
+++ b/src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
@@ -56,9 +56,14 @@ public class TileFarmStation extends AbstractPoweredTaskEntity {
       if (item == null) {
         return false;
       }
-      return match(item);
+      return match(item) && !isBrokenTinkerTool(item);
     }
-    
+
+    private boolean isBrokenTinkerTool(ItemStack item)
+    {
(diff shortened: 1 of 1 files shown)


src/main/java/crazypants/enderio/machine/spawner/ItemBrokenSpawner.java
```diff
@@ -35,14 +35,20 @@ public static String getMobTypeFromStack(ItemStack stack) {
...
   }
-  
-  public static ItemStack createStackForMobType(String mobType) {
+
+  public static ItemStack createStackForMobType(String mobType)
+  {
+    return createStackForMobType(mobType,false);
+  }
+
+  public static ItemStack createStackForMobType(String mobType, boolean wither) {
```
HenryLoenwind: mybe "boolean isVariant" instead? The "wither" is awfully specific.

hilburn: Wither Skeletons are the only entities that behave like this - in Minecraft itself or in fact any mod that I know of. If you look at other mods that have to differentiate like this, eg MopProperties: http://puu.sh/e73rn/b705442d17.png they all do the same

PR #1677, opened 2015-01-15
Process IMCs from mods who send them in PostInit

Parker8283:
There are mods that do send messages during PostInit, and are therefore not processed on EIO's side. This will allow those messages to be processed.


Parker8283:
Could I get you to build that on your Jenkins? I'm working on some integration with another mod, and would like a build to use for this.

PR #1683, opened 2015-01-15
Config Sync

hilburn:
Compresses the Config.class static (non final) values to ~1.5kB which is the lowest
I could manage without writing a whole new ByteBuf class - but not too
bad IMO as that covers about 250 syncable values including some string
arrays so a 6B/config value average is ok


CrazyPants:
Could this not be achieved in a much simpler fashion by just sending of the config from the server (stripped of comments, white space etc, then zipped), creating a temp file client side and reloading from that?
That would also ensure running a reload config command will get the servers config instead of the clients

tterrag1098:
EnderCore will contain a system for this, whether EIO will be moved to use it is questionable.

PR #1686, opened 2015-01-15
show upgrade information in Stirling generator GUI when the upgrade slot is empty

MatthiasMann:


resources/assets/enderio/lang/en_US.lang
```diff
@@ -775,6 +775,8 @@ enderio.combustionGenerator.output=Generating:
 //Stirling Gen
 enderio.stirlingGenerator.output=Generating:
 enderio.stirlingGenerator.burnRate=Burn Rate:
+enderio.stirlingGenerator.upgradeslot=Upgrade slot
+enderio.stirlingGenerator.upgrades={0} gives §f{1,number,#.#}x§r§7 efficency
```
tterrag1098: Color codes should not be included in lang files. For one, it probably will not encode correctly, and two, it is bad practice as the color is not part of the translation. You can add a `%s` or `{2}` to represent the color's position and replace it in code with an `EnumChatFormatting`.

MatthiasMann:
I moved the formatting codes into the Java code as requested

PR #1687, opened 2015-01-15
Added ComputerCraft support for PowerConduit, PowerMonitor and CapBank

Xemrox:
Added Peripheral support for ComputerCraft.
Supports most basic get/set functions.


Vexatos:
90% of these features are already in Computronics, don't you think they should be kept all in the same mod?
Also, why are you using multiple peripheral providers? A single one would have sufficed and look a lot less messy.

Xemrox:
Okay never heard of that mod. I have used the multiple providers to seperate each Peripheral completly. Anyways everything is supported then -> close

PR #1695, opened 2015-01-16
Different Item-Place-Logic

DarkStarDS9:
this fixes CrazyPants/EnderIO/#1648

the original way caused problems with items that relied on
player.inventory.currentItem, like cables from AE2 - they would set the
.currentItem-Stack to null when you placed the last cable, which in our
case would kill the pick.

Also, we’re now using the bult-in “can we place the item there?” logic
and can therefore remove canPlaceBlockOnRightClick.

It’s still not perfect though, because it doesn’t work with buckets…
see the source-code-comments. A possible solution would be to implement
a custom message that we send to ourself to switch back to the original
item slot, but I’ll just go with “you’re holding the bucket wrong”.

### DIFF
--- a/src/main/java/crazypants/enderio/item/darksteel/ItemDarkSteelPickaxe.java
+++ b/src/main/java/crazypants/enderio/item/darksteel/ItemDarkSteelPickaxe.java
@@ -5,6 +5,7 @@ import java.util.Set;
 
 import net.minecraft.block.Block;
 import net.minecraft.block.material.Material;
+import net.minecraft.client.Minecraft;
 import net.minecraft.creativetab.CreativeTabs;
 import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.entity.player.EntityPlayer;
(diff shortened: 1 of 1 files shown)


tterrag1098:
`Minecraft.getMinecraft()` on a server will crash.

DarkStarDS9:
Good call. I was checking for world.isRemote and thought that would be enough, but as it turns out it will already crash when loading the class, because client-only-classes are not loaded on a server (which makes a lot of sense ;) ). That's what you get for not testing in SMP.

Anyway, I fixed that now - this has been tested on my Direwolf20 1.7.10 server.

tterrag1098:
Egh...@SideOnly isn't the cleanest way to fix that...but ok. I need to properly clone and test this before I can merge though.

DarkStarDS9:
Well, it was used in ToolTickHandler.java for the exact same problem - so I figured it's ok ^^
What would you consider a clean way? The only other way I am aware of is a proxy-class, but that didn't feel quite right.
Anyway, you might want to add a TODO to both classes that use this, with a hint to a better solution, to discourage people from going down the same route.

PR #1723, opened 2015-01-20
Adding Fortune Machine

OldOpa:
Maybe add a expensive machine that will fortune ores inside a container (not in world like you can do with Autonomuos Activators). The crafting recipe and processing cost could be quite expensive. Also maybe you  can use enchanted books in the recipe and need special upgrades to go up to fortune three. 

Sorry for this bad pull request. First time experiencing with it.


tterrag1098:
Please submit an *issue* report, not a pull request.

Issue #746, opened 2014-06-13
Idea: Plain item conduits pull one item at a time by default


Rodg88:
Or maybe a 'speed downgrade' upgrade

tterrag1098:
Speed downgrade has been added

PR #1750, opened 2015-01-24
#746 added speed downgrade to item conduit

MatthiasMann:


src/main/java/crazypants/enderio/conduit/gui/ExternalConnectionContainer.java
```diff
@@ -193,6 +199,10 @@ public boolean canInteractWith(EntityPlayer entityplayer) {
 
   @Override
   public ItemStack slotClick(int par1, int par2, int par3, EntityPlayer par4EntityPlayer) {
+    ItemStack st = par4EntityPlayer.inventory.getItemStack();
+    if(st != null && st.getItem() == EnderIO.itemExtractSpeedUpgrade && st.getItemDamage() != 0) {
```
tterrag1098: These != 0 checks are not conducive to more types in the future. Would prefer an enum or something of the sort be created.

src/main/java/crazypants/enderio/conduit/item/ItemConduit.java
```diff
@@ -404,26 +404,20 @@ public void setOutputColor(ForgeDirection dir, DyeColor col) {
...
   public int getMaximumExtracted(ForgeDirection dir) {
-    int numUpgrades = getNumSpeedUpgrades(dir);
-    //int res = (int)Math.pow(4, numUpgrades);
-    int res = 4 + (numUpgrades * 4);
-    return res;
+    ItemStack stack = speedUpgrades.get(dir);
+    if(stack == null) {
+      return 4;
+    }
+    if(stack.getItemDamage() != 0) {
```
tterrag1098: Same here

PR #1779, opened 2015-01-25
ME counduits should now behave better

MatthiasMann:
Reworked the implementation of ME conduits - solves the following issues:
- provides player ID to be compatible with the security stations
- allows connecting and disconnecting of ME conduits using a wrench
- prevents leaking of nodes on chunk unload

Important: requires picking up existing ME conduits and placing them again to get the connections correct


tterrag1098:
With this setup: 

Right clicking the right connection does this (as expected): 

But trying to reconnect it does this (notice the back connection): 

Not sure what causes this.

tterrag1098:
It also seems unable to "merge" AE networks. If you place a controller+me conduit, add power, remove the controller and replace it, you cannot connect the conduit to the controller.

PR #1799, opened 2015-01-28
update task progress at least once per second

MatthiasMann:


tterrag1098:
What purpose does this serve?

MatthiasMann:
When processing long recipes - esp in the soul binder - the progress seems to be stuck.
For nearly all other machines/recipes (except things like dark steel ingots) the ~6% progress updates happens faster than 1 second.

The long delay between updates is noticeable now with the progress tooltips.

Issue #1123, opened 2014-10-17
Mod item Filter GUI overlap (ResourcePack Issues)


b0bst3r:
Hi

If you take an Item conduit GUI  - http://postimg.org/image/48xjkpivd/

And insert a mod item filter  - http://postimg.org/image/w7rpbkki1/

You've spliced in the area to the right of the mod name boxes (look in the bottom right corner past the Botania box).

Could you trim this area out please so from the end of the mod name boxes is the original texture?

PR #1819, opened 2015-02-01
#1123 cleanup mod item filter GUI rendering

MatthiasMann:
Makes creating textures packs for it much easier


tterrag1098:
Could I have a screenshot comparison? Or has nothing visual changed?

MatthiasMann:
Hmm, don't have one ready currently.
The text boxes got a little bit wider - they now end aligned with the player inventory.
And the inputs rows start aligned with the filter upgrade slots.

The major change is that now it is possible to retexture it for eg Sphax.

PR #1839, opened 2015-02-03
give achievements for smelting iron and cooking fish in the alloy smelter

MatthiasMann:
Also remove a redundant if statement and made 2 fields final


tterrag1098:
Isn't that method fired when the item is actually put in the slot, not when the player removes it *from* the slot? This won't work if, say, the player opens the GUI after it is done cooking, will it?

MatthiasMann:
It is exactly how the vanilla furnace works. Also that code gets triggered when you remove the item via the GUI.

PR #1857, opened 2015-02-05
drop painted glowstone when broken

MatthiasMann:


tterrag1098:
Implemented for 2.2.8 patch here: https://github.com/CrazyPants/EnderIO/commit/9f2692153794401c675cfdb05fcda8fef2cceabd

PR #1870, opened 2015-02-07
Make Photovoltaic Cell more findable

HenryLoenwind:
* Photovoltaic Cell now has a common tooltip text
* It says "Solar power" because people search NEI for "solar" not
"photovoltaic"
* Support in TooltipAddera for common tooltips of
IResourceTooltipProviders
* Fixed Advanced Photovoltaic Cell not displaying detailed tooltip text
shared with Photovoltaic Cell in Vanilla tooltip (Waila was ok)

### DIFF
--- a/src/main/java/crazypants/enderio/gui/TooltipAddera.java
+++ b/src/main/java/crazypants/enderio/gui/TooltipAddera.java
@@ -286,6 +286,20 @@ public class TooltipAddera {
     }
   }
 
+  public static void addCommonTooltipFromResources(List list, ItemStack itemstack) {
+    if(itemstack.getItem() == null) {
+      return;
+    }
+    String unloc = null;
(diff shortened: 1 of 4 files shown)


src/main/java/crazypants/enderio/gui/TooltipAddera.java
```diff
@@ -286,6 +286,20 @@ public static void addTooltipFromResources(List list, String unlocalizedName, St
     }
   }
 
+  public static void addCommonTooltipFromResources(List list, ItemStack itemstack) {
+    if(itemstack.getItem() == null) {
+      return;
+    }
+    String unloc = null;
+    if (itemstack.getItem() instanceof IResourceTooltipProvider) {
```
tterrag1098: This is now duplicated code from below. Should extract this into a private helper method that is called from multiple places.

Issue #991, opened 2014-09-08
Enhancement: Filters for vacuum chest


Xaw4:
it would be nice to have filters for the vacuum chest, just like the ones for conduits.
So the vacuum chest would only pull an item out of the world, that is "compatible" with the set filter.

would be usefull for mods that add in-world interaction.
like AE2. where you throw a Certus Quarz Seed in a paddle of water and wait till it grows. if you could set up a filter, for the vacuum chest, to only pull the grown seeds, it would make it really easy to automate.

PneumaticCraft also adds a similar method to make the PCBs.

PR #1879, opened 2015-02-08
#991 add item filters, range and redstone control and to vacuum chest

MatthiasMann:


src/main/java/crazypants/enderio/conduit/item/filter/ItemFilter.java
```diff
@@ -88,7 +92,7 @@ private boolean itemMatched(ItemStack item) {
           matched = false;
         }        
       }
-      if(!matched && isOreDicMatch(i, item)) {
+      if(!matched && useOreDict && isOreDicMatch(i, item)) {
```
tterrag1098: What does this change accomplish?

src/main/java/crazypants/enderio/machine/vacuum/TileVacuumChest.java
```diff
@@ -191,9 +226,86 @@ public boolean isItemValidForSlot(int var1, ItemStack var2) {
...
+  public boolean isItemValidForFilter(ItemStack itemstack) {
+    return itemstack != null && itemstack.getItem() == EnderIO.itemBasicFilterUpgrade && itemstack.getItemDamage() == 0;
+  }
+
+  public int getRange() {
+    return range;
+  }
+
+  private int limitRange(int range) {
+    return Math.max(1, Math.min(Config.vacuumChestRange, range));
```
tterrag1098: Maybe 0 range should be possible? Not sure how useful that would be but it seems unnecessary to floor it at 1 :P

src/main/java/crazypants/enderio/conduit/item/filter/ItemFilter.java
MatthiasMann: I moved the condition from isOreDicMatch() into itemMatched - so that it follows the same style as the other controllable filters. The item==null check in isOreDicMatch was also redundant as it is already checked in it's only caller itemMatched

src/main/java/crazypants/enderio/machine/vacuum/TileVacuumChest.java
MatthiasMann: enable / disable is handled via redstone control in the 2nd patch. Setting range to 0 would only provide a 2nd way to disable it - and also require a special check in doHover or it would waste CPU.

Terpo:
Sorry if I ask but it seems that this changes are gone in version 2.2.8.367. When I place down a vacuum chest it is simply the old one.

tterrag1098:
That build is a 2.8 release , not a 2.3 dev build. as such it does not have
the new features

Terpo:
Oh I see, sry about it^^

Issue #1843, opened 2015-02-04
Crash on placing ME Conduit (rv2)


TheJulianJES:
When I place an ME Conduit in the same block space as an existing cable, the Client crashes: 
http://pastebin.com/50fv2wFM So I have to first place the ME Conduit (Dense) and then place other conduits (ender energy) Hope I'm doing all right, also sorry for my bad English, I'm German!
Only client crash, latest AE2 Beta, latest EnderIO Jenkins. In the crashlog you can see an older version of EnderIO but it's also in the newest version :)

tterrag1098:
This doesn't seem to happen for most people. Could we have a modlist?

TheJulianJES:
It was working in a earlier version of the private "modpack", but now it isn't: http://imgur.com/a/sRNiJ
I will try it with less mods.

Akai1:
Can confirm the crash: http://pastebin.com/JUxKB3Cf

riking:
In order for the bug to trigger, `net.minecraftforge.common.DimensionManager#getCurrentSaveRootDirectory(V)Ljava/lang/File;` returns null.

MatthiasMann:
When this issue occurs - does it also affect the Dimensional Transceiver's channel list?

MatthiasMann:
@Akai1 @TheJulianJES We need a description on how to reproduce this.

TheJulianJES:
@MatthiasMann For me place an ME Conduit (Dense) into an exisiting conduit.

Akai1:
@MatthiasMann Same for me ... tried to place an ME Conduit into an Ender Energy Conduit.

MatthiasMann:
That's not what I need - I have placed tons of normal and dense ME conduits in the world and on existing conduits. All code paths for placing any kind of ME conduit will query the AE2 player data base. And that tries to get the save folder from Forge the first time it is called - which in your cases was unable to find the save game location.

Even the crash report itself could not tell you which file format version your world save uses - so something is wrong in your setup.

TheJulianJES:
It was working for me in an earlier version in AE2 OR EnderIO. Maybe the API is outdated?

TheJulianJES:
Maybe "we" could ask @thatsIch or someone from the AE2 ;)

riking:
@Akai1 @TheJulianJES Are you placing the cables in the overworld, or in another dimension?

Has the overworld ever been loaded since the server started?

Akai1:
[...]

PR #1880, opened 2015-02-08
#1843 only create conduits on server

MatthiasMann:


tterrag1098:
It might be better to do this check in the me conduit as future conduits may need this call on the client.

PR #1882, opened 2015-02-08
Portuguese Portugal translation

AnimadorX:


tterrag1098:
Isn't pt_BR still a valid language? Why rename it? I admit that I might not know all the caveats of the MC language system.

AnimadorX:
Well, it isn't to rename it. Is to add. I'm very confusing with mod
creation and more stuff.

>

-- 
Enviado do meu iPad

tterrag1098:
Your PR renames the file to pt_PT, instead of copying it and adding a new file.

AnimadorX:
Yes.... I know that. Github is new for me (so it is confusing)

-- 
Enviado do meu iPad

tterrag1098:
Ok well, it's easy to redo this in the right way. When you go to the lang folder there should be a "plus" (:heavy_plus_sign:) button at the top, right at the end of the file path . Then just copy the existing pt_BR into your new file and name it pt_PT.lang and you're done :P

tterrag1098:
@Adaptivity implemented your changes above

PR #1897, opened 2015-02-10
Optimizations for doInsertItem()

HenryLoenwind:
Logic is the same as my last implementation, it is just heavily
optimized. As it may be called many times per tick in big networks doing
that looked appropriate.

Note: One of the optimizations relies on the fact that the code only
runs single-threaded in the server thread.

### DIFF
--- a/src/main/java/crazypants/util/ItemUtil.java
+++ b/src/main/java/crazypants/util/ItemUtil.java
@@ -76,59 +76,58 @@ public class ItemUtil {
    * Insert items into an IInventory or an ISidedInventory.
    */
   private static int doInsertItemInv(IInventory inv, ItemStack item, ForgeDirection inventorySide) {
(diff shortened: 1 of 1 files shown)


kayila:
If it relies on being single threaded, might it be a good idea to use a singleton or a mutex lock to ensure that, should that piece of code ever get pushed to multithreaded, it doesn't choke and die?

MatthiasMann:
The way inventories are implement does not even allow to multi thread that code. It would require that all inventory operations in the game are either atomic or transactions.

HenryLoenwind:
@MatthiasMann Exactly. I just wanted to point out that I used a construct that's an anti-pattern for Java coding on purpose.

Third way to  introduce multi-threading for inventories would be getLocks(IInventory ...) with some insane order-independent no-deadlocking algo. For Minecraft that might be the better route to take ;)

PR #1924, opened 2015-02-14
save locked slots state when picking up a Farming Station

MatthiasMann:
Also cleanup GUI code and locked state handling


src/main/java/crazypants/enderio/machine/farm/GuiFarmStation.java
```diff
@@ -30,25 +33,23 @@ public void initGui() {
...
+    return new ToggleButtonEIO(this, LOCK_ID+slot, x, y, IconEIO.FARM_UNLOCK, IconEIO.FARM_LOCK).setSelected(getTileEntity().isSlotLocked(slot));
   }
 
   @Override
   protected void drawForegroundImpl(int mouseX, int mouseY) {
     super.drawForegroundImpl(mouseX, mouseY);
 
-    if(inventorySlots.inventorySlots.size() >= getTileEntity().maxSupSlot && !isConfigOverlayEnabled()) {
-      for (int i : getTileEntity().lockedSlots) {
-        if (i < inventorySlots.inventorySlots.size()) { // hack to allow old broken farms to be opened
```
tterrag1098: Will this still be compatible with farms with corrupted data?

MatthiasMann: Yes, as it now only iterates over the supply slots

PR #1926, opened 2015-02-14
Removed unnecessary client sync

HenryLoenwind:
50% FPS drop from just a single buffer is a bit too much...


tterrag1098:
How exactly was the previous code flawed?

HenryLoenwind:
@MatthiasMann Could you elaborate?

MatthiasMann:
Returning true from processTasks will cause a block update between server and client. But there is already a lower frequency sync of the energy buffer in the base class which works without a block update. So returning false saves on block updates without causing a desync.

HenryLoenwind:
Thanks, MatthiasMann.

I'd like to add that we found this issue because placing a single energy buffer cause me to lose about 50% FPS. It's very repeatable by switching between an energy conduit and a buffer. (MP with server in same LAN on a different PC).

PR #1927, opened 2015-02-14
Dark Steel Shears

HenryLoenwind:
* Dark Steel Shears are crafted with 2 Dark Steel Ingots
* They are configurable more durable and configurable faster
* They can be empowered and will then AOE with a configurable radius
* I plan on adding another upgrade later (diamond edge)


ChatFawkes:
A suggestion for a future upgrade if possible (and accepted by the devs):

An upgrade similar to the Rainbow upgrade for Dartcraft Force Shears. The Rainbow upgrade allowed for random colored wool drops from sheep regardless of their color.

HenryLoenwind:
If you have an idea for a technological(!) reason for the wool to turn color, I could add that.

tterrag1098:
Please in the future avoid the whitespace changes, it's very hard to see what was changed.

Issue #1937, opened 2015-02-16
Me conduits and dark pick bugs.


deadgamer1030:
Me conduits do not link with p2p tunnels properly after logging back into a world, they disconnect.
Dark pick with the spoon upgrade does not work on Tic ore gravels. It will break the gravel but not drop anything.

6210classick:
provide more information please like what Forge & EIO & AE are you using ?

EIO ME Conduit do not interact with any AE2 buses directly you need to use AE2 cable in between also be aware that in almost every version of EIO from 2.2.0 ME Conduits causes alot of problems and crashes its even can corrupt your world permanently ,, couple days ago in version 2.2.8.349 supposedly all ME Conduits related problems has been fixed

tterrag1098:
I would also be interested in the version because I am fairly sure the gravel ores should work.

deadgamer1030:
version 2.2.8.349

PR #1941, opened 2015-02-16
#1937 reconnect to P2P ME tunnels after chunk load

MatthiasMann:
Sadly I found no better solution - could be cleaned up once someone either fixes the AE2 API or finds it's magic decoder ring


MatthiasMann:
I updated the commit - seems like everything works now except when the QNB is formed / broken as that is missing a neighbor block update from AE2.

tterrag1098:
Screw everything about this

PR #1956, opened 2015-02-18
Added missing SideOnly annotations

HenryLoenwind:
* Added @SideOnly(Side.CLIENT) to methods that override methods that are
@SideOnly(Side.CLIENT).
* Added @SideOnly(Side.CLIENT) to classes that extend classes that are
@SideOnly(Side.CLIENT)
* Added @SideOnly(Side.CLIENT) to ClientProxy class

Note: These SideOnlys are useful because they make it clear for a programmer which methods he cannot call from server-side code. They cannot be used because the jvm doesn't like it at all to execute a method/class that overrides/extends a non-existing methods/class.

This change is harmless. If any of the newly annotated methods would have been called server-side, the game would have crashed hard in the past, too.


HenryLoenwind:
sorry, still had "cleanup imports" and "strip whitespace at end of line" on.

PR #1964, opened 2015-02-18
Prevent Non-Active blood magic tools being used

hilburn:
It's not completely accurate as it's feasible someone else has an axe
with a boolean tag "isActive" - but I think it's likely the station
shouldn't use those either


hilburn:
actually it was a bit cleaner to fix this on the BM side, never mind

PR #1966, opened 2015-02-18
Split off client side GuiHandlers

HenryLoenwind:
* All getClientGuiElement() methods are now client side only
* The GuiScreens they use, too (they extend vanilla's sided GuiScreen)
* Had to introduce our own version of IGuiHandler, "ISidedGuiHandler"
because fml's one mixes sides
* Split our central GuiHandler into client and server and moved it to the proxy
* Missed a couple of getIcon()s last time

Note: I'm still playing it save---only code that would have crashed and
burned server-side is annotated.

### DIFF
--- a/src/main/java/crazypants/enderio/ClientGuiHandler.java
+++ b/src/main/java/crazypants/enderio/ClientGuiHandler.java
@@ -1,0 +1,25 @@
+package crazypants.enderio;
+
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.world.World;
+import cpw.mods.fml.relauncher.Side;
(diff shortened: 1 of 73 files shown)


CrazyPants:
I am not sure what the motivation for this is. I am not aware of it ever having caused an issue as is, and it certainly complicates things.

HenryLoenwind:
This was part of a cleanup to properly mark things a client-only to avoid people calling those methods my accident.

CrazyPants:
Given the complexity this adds, I dont think this is good change. People can still call methods by accident, and they will find out pretty quickly anyway.
Thanks for the effort though. Much appreciated.

PR #1980, opened 2015-02-21
Create crash

alegal200:


MatthiasMann:
What is this bullshit? Just create a issue like every sane person would - and don't create a pull request!

PR #1992, opened 2015-02-23
New gradle task to help translators

HenryLoenwind:
* The new gradle task "checkTranslations" creates reports on
missing/extra lang file entries for all language files
* It is not included in the normal build process and must be run by hand
* The reports are gitgnore'ed

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -167,3 +167,35 @@ curse {
     changelog = System.getenv("CHANGELOG").equals("none") ? getChangelogText() : System.getenv("CHANGELOG")
     releaseType = project.hasProperty('mod_appendix') ? "${mod_appendix}".toString() : 'release'
 }
+
+task checkTranslations << {
+    Map<String, String> mapen = new HashMap<String, String>()
+    (new File('resources/assets/enderio/lang/en_US.lang')).eachLine {
+        def (value1, value2) = it.tokenize( '=' )
+        if (value1 == null || value2 == null) {return}
+        mapen.put(value1, value2)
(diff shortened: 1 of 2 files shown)


HenryLoenwind:
Yes, there are tools for working with .lang files. But this is about as low on skill requirements as possible. ;)

build.gradle
```diff
@@ -167,3 +167,31 @@ curse {
...
+
+task checkTranslations << {
+    Map<String, String> mapen = new HashMap<String, String>()
+    (new File('resources/assets/enderio/lang/en_US.lang')).eachLine {
+        def (value1, value2) = it.tokenize( '=' )
+        if (value1 == null || value2 == null) {return}
+        mapen.put(value1, value2)
+    }
+    
+    for ( lang in ['cs_CZ', 'de_DE', 'fr_FR', 'ko_KR', 'pl_PL', 'pt_BR', 'ru_RU', 'zh_CN', 'zh_TW'] ) {
```
Vexatos: Couldn't you dynamically check for files inside /lang ?

HenryLoenwind: I decided it to be not worth the effort. It's not like we get new translation every week or so...

(And honestly, I was tired of looking up Groovy syntax and examples. I'm sure there still are plenty of Javaisms in there that could be written in a simpler and more precise way...)

tterrag1098: I would much prefer a wildcard *.lang selector instead of hardcoded names, I know we don't get new translations often, but when we do it'll break.

PR #2051, opened 2015-03-04
Fix Mouse Wheel Handling

### LINKED ISSUE #2049: Mouse Wheel Scroll
enderio is blocking all other mods from using Mouse Scroll Wheel? there are other mods that cant use the scroll wheel because enderio. is there something that can be done?

HenryLoenwind:
* Mouse wheel data is no longer consumed unless acted upon; fixes #2049
* Mouse wheel handling re-coded to make sense


HenryLoenwind:
There's a better way to do this. #2052

Issue #2096, opened 2015-03-17
Farming Station drops items on the ground


emperor06:
When a farming station harvests too many items and overflows its internal item buffer, items are dropped on the ground leading to potential server lags and item loss.
This happens quite often when farming trees with shears on as a tree can produce quite a large number of items (wood, leaves, fruits) or when a slow speed pipe is used to extract the items from the station.
A vacuum chest can be used as a workaround tho.

HenryLoenwind:
This is the intended behavior. However, it may be a good idea to extend the internal storage, now that we shear leaves. Goes to my TODO list.Thanks for the heads up.

roboderpy:
Still seems to be a problem.
![2016-09-09_13 30 30](https://cloud.githubusercontent.com/assets/6339511/18377637/6e073e10-7692-11e6-956c-e1a2099fad40.png)
![2016-09-09_13 38 32](https://cloud.githubusercontent.com/assets/6339511/18377638/6e2cefa2-7692-11e6-82cb-0ac85d200cd6.png)
Basically, farm station should try to push items immediately into configured side when it's internal inventory overflows when it chops trees

PR #2101, opened 2015-03-17
Add more output slots to farming station

HenryLoenwind:
* Added 2 output slots to farming station to prevent overflow from
shearing leaves
* Resized GUI to make space
* Solves #2096


src/main/java/crazypants/enderio/machine/farm/FarmStationContainer.java
```diff
@@ -9,6 +9,18 @@
 
 public class FarmStationContainer extends AbstractMachineContainer {
 
+  private static final int ROW_TOOLS = 19;
+  private static final int ROW_IO = 44;
+
+  private static final int COL_TOOLS = 44;
+  private static final int COL_INPUT = 53;
+  private static final int COL_OUTPUT = 116;
```
tterrag1098: Could you move the output slots over half a slot width? So that they are centered on where they used to be. I think it looks weird that close to the edge. Just a nit-pick.

HenryLoenwind: no problem

src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
```diff
@@ -643,13 +655,23 @@ public void readCommon(NBTTagCompound nbtRoot) {
...
           inventory[i] = null;
         }
       }
+      slotLayoutVersion = 1;
+    }
+    if (slotLayoutVersion < 2) {
+      // update from v1 (3+4+4+1) to v2 (3+4+6+1) by moving cap to new position
+      if (inventory[(3+4+6+1)-1] == null) {
+        inventory[(3+4+6+1)-1] = inventory[(3+4+4+1)-1];
+        inventory[(3+4+4+1)-3] = null;
```
tterrag1098: Why not just use getSlotDefinition().getMaxOutput() ?

HenryLoenwind: because that only gives the current implementation. once the layout updates again, it won't give good data. however, that code already is outdated, see latest PR

PR #2126, opened 2015-03-20
Added bone meal support to farming station

HenryLoenwind:
* 2 new slots for bone meal
* Will try to use it on everything thats on the field and cannot be
harvested
* Different RF usage for successful and unsuccessfull tries
* Fix: Output slots no longer accept seeds
* Fix: "Only one tool per slot" policy works again

(Note: This PR contains the commits of the other two PRs for now. I'll update it once they have been pulled.)


tterrag1098:
Done

src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
```diff
@@ -424,23 +450,50 @@ protected void doTick() {
...
-        if(harvest.getDrops() != null) {
-          PacketFarmAction pkt = new PacketFarmAction(harvest.getHarvestedBlocks());
-          PacketHandler.INSTANCE.sendToAllAround(pkt, new TargetPoint(worldObj.provider.dimensionId, bc.x, bc.y, bc.z, 64));
-          for (EntityItem ei : harvest.getDrops()) {
-            if(ei != null) {
-              insertHarvestDrop(ei, bc);
-              if(!ei.isDead) {
-                worldObj.spawnEntityInWorld(ei);
-              }
+      if(harvest != null && harvest.getDrops() != null) {
```
tterrag1098: I get what you did here, but it makes it very difficult to see what changed. I'd prefer if you didn't change indentation unless necessary.

src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
```diff
@@ -321,21 +347,21 @@ protected boolean isMachineItemValidForSlot(int i, ItemStack stack) {
...
+          if (ToolType.getToolType(stack).itemMatches(inventory[j])) {
+            return false;
           }
-          return true;
         }
+        return true;
       }
       return false;
+    } else if (i <= maxFirtSlot) {
+      return stack.getItem() == Items.dye && stack.getItemDamage() == 15;
```
tterrag1098: It would be nice if there was a configurable list for this. Being able to use Forestry/MFR fertilizer would be great.

src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
HenryLoenwind: Leaving code at the wrong indentation strikes me as a "no go"...

tterrag1098: I meant to just leave it alone ;)

src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
HenryLoenwind: I'm not sure on how to apply the other fertilizers (SFX/no SFX). I'd like to investigate that first. And then a list would not be that helpful, just as with the tools you need to detect every item differently.

tterrag1098: Alright, we'll save that for another time.

HenryLoenwind: Short Google check says that MFRs Industrial Fertilizer does not support right-clicking by hand. So it'd need special coding anyway.Forestry also looks like it needs machine application. IC2 might work directly.

Issue #2128, opened 2015-03-20
SMP Server crash at start (NoClassDefFoundError EntityClientPlayerMP)


emperor06:
On commit 6d8cf393d9bd4612f640ec2a8558c535e493f26a
The client starts just fine but the server crashes. 

[Server thread/ERROR] [FML]: Caught exception from EnderIO
java.lang.NoClassDefFoundError: net/minecraft/client/entity/EntityClientPlayerMP

Full log here:
http://pastebin.com/jKYgcXe2

Kodos-Atoz:
Do you by chance use Galacticraft?

emperor06:
Nope.
It seems to be related to magnet/bauble. I use Thaumcraft which relies on bauble too if I remember correctly. I don't know if that's the cause.

Kodos-Atoz:
Hm, can you pastebin your full log?

emperor06:
There you go:
http://pastebin.com/C6qpQQEt

emperor06:
If I had @SideOnly(Side.CLIENT) to MagnetController.handleKeypress(), the server is happy. But I don't know if the whole method was intended to run only on client side.

HenryLoenwind:
Trusting the IDE to auto-set the type of a local variable correctly *usually* works. Unless it's Forge...

Could you please test if this works or if now the Yeta wrench crashes because the KeyTracker is Client-only?

emperor06:
Sure, testing now

emperor06:
As far as I can tell:
- no crash at start on server nor client
- no crash neither in single player
- no crash with the Yeta wrench
- Yeta wrench can wrench and its key binding works perfectly
- magnet activates properly when the key binding is used

The only thing that doesn't seem to work is I can't slot the magnet in my bauble slots (or maybe that's intended?)

tterrag1098:
The key tracker should be a client side class. What is loading it on the
server?

HenryLoenwind:
It wasn't the KeyTracker, it was the key handling method in the Commander, that used a client-only player class as parameter. Crash on classloading. But I added a clientonly annotation while fixing it on the keyhandler, and there's a method in the yeta that references the keyhandler. I was not 100% sure if that would crash on classloading (happening on server) or execution (not happening ob server). It is the later, so we're good.

PR #2130, opened 2015-03-20
Fix: Sided Code

HenryLoenwind:
re #2128


tterrag1098:
The use of SideOnly is discouraged on non-vanilla methods and classes. Fixed an alternate way in af4c7ce5870a116ee85791bb321664eccb2f4dbd

PR #2136, opened 2015-03-21
Added flower support to farming station

HenryLoenwind:
* Farming station will now pick flowers (not plant!)
* Supported flowers: Vanilla single block, BoP, Botany, Botania

### DIFF
--- a/src/main/java/crazypants/enderio/machine/farm/FarmersRegistry.java
+++ b/src/main/java/crazypants/enderio/machine/farm/FarmersRegistry.java
@@ -22,6 +22,7 @@ public final class FarmersRegistry {
     addIC2();
     addMFR();
     addThaumcraft();
+    addFlowers();
 
     FarmersCommune.joinCommune(new StemFarmer(Blocks.reeds, new ItemStack(Items.reeds)));
     FarmersCommune.joinCommune(new StemFarmer(Blocks.cactus, new ItemStack(Blocks.cactus)));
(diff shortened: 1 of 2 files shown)


src/main/java/crazypants/enderio/machine/farm/FarmersRegistry.java
```diff
@@ -191,6 +192,16 @@ private static void addExtraUtilities() {
     }
   }
 
+  private static void addFlowers() {
+    FarmersCommune.joinCommune(new FlowerPicker().add(
+        GameRegistry.findBlock("minecraft", "yellow_flower"), 
+        GameRegistry.findBlock("minecraft", "red_flower"),
```
tterrag1098: Why? lol

HenryLoenwind: Number of reasons:

(a) because I can
(b) because I don't want bug reports about "eio doesn't harvest flowers but it's in the changelog"
(c) because there's a change they get spawned by using bonemeal if the user has a block of grass where there should be a plant. (Tall grass already gets replaced, so that's no issue.)

tterrag1098: No I meant why are you using `GameRegistry.findBlock`...

HenryLoenwind: Oh, that.

Nicer pattern with the rest, easier to move to a config file in the future.

PR #2143, opened 2015-03-22
The Mighty Powered Drain

HenryLoenwind:
A new machine that picks up liquids from the world.

Should the recipe include a head?

Works best with it the core mod Water Hooks (https://github.com/HenryLoenwind/waterhook) installed, but doesn't require it.

Cannot be crafted on dedicated servers by default (config option),


src/main/java/crazypants/enderio/machine/drain/FluidHelper.java
```diff
@@ -0,0 +1,328 @@
...
+      if (isSourceBlock(world, direction)) {
+        Fluid fluidForBlock = FluidRegistry.lookupFluidForBlock(direction.getBlock(world));
+        if (fluidForBlock != null) {
+          FluidHelper result = getInstance(world, new FluidStack(fluidForBlock, 1000), direction);
+          if (result != null) {
+            return result;
+          }
+        }
+      }
+    }
```
CrazyPants: Why are you running the loop twice?

src/main/java/crazypants/enderio/machine/drain/FluidHelper.java
```diff
@@ -0,0 +1,328 @@
...
+      return false;
+    }
+  }
+
+  private static final ForgeDirection[] DIRECTIONS_INIT = {ForgeDirection.UP, ForgeDirection.NORTH, ForgeDirection.SOUTH, ForgeDirection.WEST, ForgeDirection.EAST};
+  
+  public static FluidHelper getInstance(World world, BlockCoord bc) {
+    for (ForgeDirection forgeDirection : DIRECTIONS_INIT) {
+      BlockCoord direction = bc.getLocation(forgeDirection);
+      if (isSourceBlock(world, direction)) {
```
CrazyPants: Why this check and not just != null, they both do the same thing anyway

CrazyPants:
How well does it work without the core mod? Is it stable enough / performs well with out it?
Sound like a cool addition.

src/main/java/crazypants/enderio/machine/drain/FluidHelper.java
HenryLoenwind: The logic is to first look for a source block, but if there's none to take a flowing block, too.

I could fold them into one loop like

    foreach (EPass pass : EPass.values()) {
        ....
        if (pass == EPass.PASS2 || isSourceBlock(world, direction)

but for just 2 passes and a couple of lines of code it looks like a big overkill. (And the reverse of the kind of loop unrolls any good optimizer does ;))

src/main/java/crazypants/enderio/machine/drain/FluidHelper.java
HenryLoenwind: Sorry, I'm not 100% sure which of the lines that Github shows me you commented on.

The isSourceBlock() certainly is no simple null check; code is just above.

HenryLoenwind:
Works fine here. I put it under some heavy stress testing (ocean draining) without issues. I tested it with vanilla and ender io fluids, and some modded ones I had in my dev env anyway.

Without the core mod, it will act as an infinite water source, no breakage. Just way more cpu usage than a Reservoir... ;)

What makes it heavy is vanilla's heavy liquid flowing code, or with Lava the heavy re-lighting.

tterrag1098:
After some testing I found some annoyances in how it functions.

Mainly the "pulling" of sources towards it, while it is a bit neat, it is unnecessary lag for little aesthetic gain.

Finally, it doesn't seem to properly drain even "natural" lakes, which would be very annoying in gameplay.

tterrag1098:
I am planning to do "the big merge" of the EnderCore branch soon, which means you will have to do a lot of work to sync up. If those are fixed soon I can pull this maybe.

tterrag1098:
I am back from vacation, and nothing has been done on this, so I am putting it on hold so I can move forward with other things.

HenryLoenwind:

HenryLoenwind:
This is currently live in EnderIOAddons, if I bring this over here it'll be in a new branch.

PR #2153, opened 2015-03-24
Added magicalcrops_MagicalCropFertilizer

HenryLoenwind:


Adaptivity:
What about MineFactoryReloaded Industrial Fertilizer?

HenryLoenwind:
@Adaptivity If it cannot be right-clicked on a crop it has to wait. I'll do those later, because I have to find out their "potency" in comparison to bone meal.

No sense in wasting a super-duper-ultra fertilizer that takes 5 dragon eggs to craft and can fertilize whole dimensions to grow a single wheat by one phase. ;)

tterrag1098:
Please go with a smaller enum name...lol

HenryLoenwind:
I was tempted to use "M", but couldn't bring myself to do so. ;)

PR #2176, opened 2015-03-29
Add our take on unsupported mods to the crash report

HenryLoenwind:
Also stop early when AE2 rv1 is loaded

### DIFF
--- a/src/main/java/crazypants/enderio/tool/EnderIOCrashCallable.java
+++ b/src/main/java/crazypants/enderio/tool/EnderIOCrashCallable.java
@@ -1,0 +1,56 @@
+package crazypants.enderio.tool;
+
+import java.util.List;
+
+import cpw.mods.fml.client.FMLClientHandler;
+import cpw.mods.fml.common.FMLCommonHandler;
+import cpw.mods.fml.common.ICrashCallable;
+import cpw.mods.fml.common.Loader;
+import cpw.mods.fml.common.ModAPIManager;
+import cpw.mods.fml.common.ModContainer;
+import cpw.mods.fml.relauncher.Side;
+import crazypants.enderio.EnderIO;
+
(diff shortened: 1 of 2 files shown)


tterrag1098:
Why not use the @Mod dependencies field?

src/main/java/crazypants/enderio/tool/EnderIOCrashCallable.java
```diff
@@ -0,0 +1,69 @@
...
+    if (hasOptifine) {
+      msg += "\"Optifine is installed. This is NOT supported.\" ";
+    }
+    if (hasBadAE2) {
+      msg += "\"AE2 rv1 is installed. Ender IO is no longer compatible with rv1, please update to rv2.\" ";
+    }
+    if (hasUnknownAE2) {
+      msg += "\"An unknwon AE2 API is installed. Ender IO was build against API version rv2.\" ";
+    }
+    msg += "This may have caused the error. Try reproducing the crash WITHOUT this/these mods before reportin it.";
```
Vexatos: "reportin"........

HenryLoenwind:
Ugh. Never directly PR a change that was done in one sitting after a very long day at work...

Sorry, quality was bad, feature set was bad. This version should be more sensible in both.

tterrag1098:
I'm ok with the Opifine bit, but AE2 is no longer needed.

HenryLoenwind:
I already removed the rv1 check...

CrazyPants:
Is optifine really needed? I havent been around for a bit but I cant remeber any crash bugs being reported that where caused by optifine. Any issues I have seen with it where rendering related. I am not 100% sure if it these even effected eio.

HenryLoenwind:
We had some "already tessellating" ones, I think.

But this PR is more or less becoming a "add the class so we can add messages later if we need to"...

PR #2206, opened 2015-04-01
Add Thaumaturge's Armor vis discount upgrades.

eduardog3000:
Add Thaumaturge's Armor vis discount upgrades for Dark Steel Armor.


eduardog3000:
Actually, let me add some localization in other languages first.

eduardog3000:
Alright, I copied localizations for "Thaumaturge's [Robe/Leggings/Boots]" from Thaumcraft, and copied the localization for "Vis Discount" from the tooltip of the Revealing upgrade.

tterrag1098:
Why add duplicate lang entries instead of just using Thaumcraft's? We know it is installed when this is active.

eduardog3000:
Hmm, I didn't even think of that. I'll change that and rebase.

eduardog3000:
Done, but the tooltip has to be `item.ItemChestplateRobe.tooltip.detailed.line1` rather than following the standard `enderio.darksteel.upgrade.thaumaturgeArmor.chestplate.tooltip.detailed.line1`.

tterrag1098:
I think this is better left out, we don't want to completely re-implement TC's armor, EIO armor can do enough as it is.

PR #2232, opened 2015-04-04
Bread Production

Dwargkiller:
Added Flour, altered Wheat's sag mill recipe, added Flour->Bread alloy
smelter recipe.

Added the flour to avoid any dependencies on other mods. 

The main reason behind the changes is to facilitate the increase of bread yield from wheat. The usual output of a grinder/mill when given wheat is flour (both realistically and in other mods), and throwing wheat into your sag mill and getting seeds can be jarring.This fixes that entirely!


tterrag1098:
Hate to nitpick, but the item would probably fit better as part of the `PowderIngot` enum, not `Material`. This would also avoid it being automatically registered as "itemFlour"

Dwargkiller:
Keep nitpicking, I'll adjust and get back to you.

PR #2239, opened 2015-04-05
Spawners get nasty when broken

HenryLoenwind:
At least a little bit


tterrag1098:
What does this fix?

HenryLoenwind:
Did I tag this "fix"? Nope, it's an enhancement.

Reasoning: People are more and more doing "break spawner fast, then deal with mobs" instead of the traditional "light up area, kill mobs, break spawner". And as EIO already meddles with spawners...

tterrag1098:
Oh! It spawns mobs when it's broken. I like that, yes.

HenryLoenwind:
Oh, yes. That method is not obvious from its name

PR #2251, opened 2015-04-06
Recipe cleanup

Dwargkiller:
Cleaned up outdated code and used GameRegistry instead. Also fixed a
small spelling error.

### DIFF
--- a/src/main/java/crazypants/enderio/material/MaterialRecipes.java
+++ b/src/main/java/crazypants/enderio/material/MaterialRecipes.java
@@ -12,7 +12,6 @@ import net.minecraft.init.Blocks;
 import net.minecraft.init.Items;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
-import net.minecraft.item.crafting.FurnaceRecipes;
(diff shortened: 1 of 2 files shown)


src/main/java/crazypants/enderio/material/MaterialRecipes.java
```diff
@@ -109,7 +108,7 @@ public static void addRecipes() {
     } else {
       GameRegistry.addShapedRecipe(cbc, "ggg", "scs", "ggg", 'g', Blocks.gravel, 's', Blocks.sand, 'c', Items.clay_ball);
     }
-    FurnaceRecipes.smelting().func_151394_a(binderComposite, conduitBinder, 0);
+    GameRegistry.addSmelting(binderComposite, conduitBinder, (float) 0.5);
```
tterrag1098: Why increase the xp amount?

Dwargkiller: Conforms to vanilla XP amounts, rather than have none entirely. Can be changed if wanted.

PR #2253, opened 2015-04-06
Add Natura Barley to Flour SAG Mill recipe.

eduardog3000:
Expanding on #2243, barley can now be SAG milled into flour.

### DIFF
--- a/resources/assets/enderio/config/SAGMillRecipes_Core.xml
+++ b/resources/assets/enderio/config/SAGMillRecipes_Core.xml
@@ -2459,6 +2459,19 @@ addOreDictionaryTooltips=true
     </recipe>
   </recipeGroup>
 
+  <recipeGroup name="Natura">
+
+    <recipe name="Barley" energyCost="800">
+      <input>
+        <itemStack modID="Natura" itemName="barleyFood" />
+      </input>
+      <output>
+        <itemStack modID="EnderIO" itemName="itemPowderIngot" itemMeta="8" number="2" />
+        <itemStack modID="Natura" itemName="barley.seed" chance="0.2" />
+      </output>
+    </recipe>
+  </recipeGroup>
+
   <grindingBalls>
     <grindingBall id="Flint" grindingMultiplier="1.2" chanceMultiplier="1.25" powerMultiplier="0.85"
       durationRF="24000">


Adaptivity:
Nobody registered barley in OreDict expect MatrexsVigil, I guess. I requested this years ago for Natura and Biomes O Plenty, still no progress. So, this probably impossible.

Dwargkiller:
Vexatos: You can craft Barley into 1 natura flour. This would let you sag mill barley into 2 EIO flour.

Vexatos:
@Dwargkiller That'd be pretty unbalanced; why should that work? Applied Energistics flour is made in an AE2 grindstone (i.e. pulverizer with a hand crank) and still only gives one piece of flour.

Dwargkiller:
Wheat can be pulverized in a pulverizer into 2 Natura flour. Wheat can currently be sag milled into two EIO flour, that are smelted at a 1:1 ratio into bread. This change would bring Barley in line with wheat in terms of bread production.

resources/assets/enderio/config/SAGMillRecipes_Core.xml
```diff
@@ -2459,6 +2459,19 @@ addOreDictionaryTooltips=true
...
   </recipeGroup>
 
+  <recipeGroup name="Natura">
+
+    <recipe name="Barley" energyCost="800">
+      <input>
+        <itemStack modID="Natura" itemName="barleyFood" />
+      </input>
+      <output>
+        <itemStack modID="EnderIO" itemName="itemPowderIngot" itemMeta="8" number="2" />
```
tterrag1098: The benefit here is automation, not necessarily the output bonus. Maybe make this 1 with a 50% chance of an extra flour?

eduardog3000: Right now wheat gives 2, and since other mods treat barley and wheat the same, so should EnderIO.

Lilylicious: The whole barely/wheat grinding bonus things is a sticky mess, honestly. Here's the current status:

Natura barley crafts in a crafting table into one flour.
Thermal Expansion machines grind barley and wheat, with a 1:2 ratio of input to output.

There's not many other popular mods that aren't spiraling downwards in a fiery circle of death that do anything with wheat. Do we want to follow the thinking in TE, giving a bonus to outputs? Or do we want to just do automation of wheat->flour? To be fair, flour is already a triple on bread output per input compared to vanilla wheat.

One option is to keep the current recipe in this PR as is, and come back to this as well as our current wheat->flour recipe in a balancing mood, where we decide if we want all wheat and wheat-like products to give 1:1(or 1:1.5 (1:1.25 is another option)) wheat or if we want a steady 1:2 output.  That way we do balancing of flour in general in one swoop rather than balance this now and do the other flour later.

CrazyPants: I will merge it as is so its the same as wheat. I think it might be better balanced at 1:1.5, but food / bread production is not really much of a blanace issue either way as it is so easy to get and not that useful anyway regardless of the decision here.

PR #2261, opened 2015-04-07
Nether Quartz Balance

Dwargkiller:
Mining nether quartz without luck gives 1 quartz per ore. Comparable
mods give two per grinding. Certus Quartz gives one crystal one dust in
the sag mill recipe.

This change makes Nether Quartz comparable to Certus Quartz in EIO
recipes, and comparable to other mods, while still improving compared to
vanilla yields (just not quadruple)


tterrag1098:
I'd prefer if it was just 2 quartz, with may be a chance of dust. I think I will implement this myself. Thanks for the heads-up on the imbalance though.

PR #2267, opened 2015-04-09
Typo  fixup

RAnders00:
Should be easy to merge this.


eduardog3000:
It's still spelled wrong, it is "transceiver". And it should probably capitalized, since it is part of "Dimensional Transceiver".

eduardog3000:
Better, but you should probably squash those into one commit.

Issue #307, opened 2013-11-20
SAG Recipes for Aluminum/Aluminium .


Unrepentant-Atheist:
A few mods add the following to the ore dictionary with their respectives ores and whatnot:

dustAluminum

ingotAluminium
ingotAluminum

oreAluminum
oreNaturalAluminum

Would be nice if the SAG could handle it and have it's own dust for it like tin, copper, etc.

PR #2279, opened 2015-04-11
Aluminum Support

Dwargkiller:
Sag Mill Aluminum oredict support

### DIFF
--- a/resources/assets/enderio/config/SAGMillRecipes_Core.xml
+++ b/resources/assets/enderio/config/SAGMillRecipes_Core.xml
@@ -590,6 +590,24 @@ addOreDictionaryTooltips=true
         <itemStack oreDictionary="dustPlatinum" chance="0.1" />
       </output>
     </recipe>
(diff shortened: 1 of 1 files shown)


Adaptivity:
What about Aluminium? It is same to Aluminum but some mods uses Aluminium name.

tterrag1098:
@Adaptivity Then those mods are wrong. Ore dictionary is done in american english, that's the standard.

Vexatos:
@tterrag1098 Not really, considering the proper scientific term is Aluminium, even in American English. Aluminum is just the more common term there.

tterrag1098:
@Vexatos regardless of technicalities, the majority of mods already use "um", so I see no need to support doube-standards.

Adaptivity:
@tterrag1098 there are an issue open for that: https://github.com/SleepyTrousers/EnderIO/issues/307 so it still might worth adding Aluminium.

PR #2290, opened 2015-04-12
Update Forge and fix Fluid API

wasweb:
It should be done sooner or later. 
Untested!

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,15 +1,15 @@
 minecraft_version=1.7.10
-forge_version=10.13.2.1291
+forge_version=10.13.3.1370-1.7.10
 
 mod_version=2.3.0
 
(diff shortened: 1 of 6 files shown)


MatthiasMann:
Why did you make an untested PR? Just test it before creating the PR.

wasweb:
I mean it's working for me. I just warn you guys to check it out before merging it.

tterrag1098:
I would rather not force a forge update, we currently only require a very old version of forge, and I see no reason to make this change.

TheJulianJES:
@tterrag1098 Since it will crash with newer forge versions.

tterrag1098:
@TheJulianJES No it won't.

TheJulianJES:
@tterrag1098 O_o

Issue #2296, opened 2015-04-12
Crouch to activate travel anchor option


Xuerian:
A simpler version of/precursor to #1140

Once you have a few items that enhance jump capacity or add flight, jumping to travel can get a little wonky.

6210classick:
it's not a bug it's just the dafault behavior , i believe the same thing also happens with open blocks elevators

Xuerian:
@6210classick I know, this was a feature suggestion. GitHub doesn't really let issue creators not on the project label issues, it'll get labeled by someone.

PR #2301, opened 2015-04-12
Elevator functionality and sneak activation

Lilylicious:
Solves the "issue" in #2296 where jump enhancements/flight can cause difficulties in activating travel anchors and/or cause increased iteration periods between attempts.

### DIFF
--- a/resources/assets/enderio/lang/en_US.lang
+++ b/resources/assets/enderio/lang/en_US.lang
@@ -879,6 +879,9 @@ enderio.gui.travelAccessable.protected=Protected
 enderio.gui.travelAccessable.unauthorised=You are not allowed to access this block.
 enderio.gui.travelAccessable.enterCode=Access Code Required
 enderio.gui.travelAccessable.ok=Authorise
+enderio.gui.travelAccessable.skipLocked=Skipped locked anchor, use targeted teleport.
+enderio.gui.travelAccessable.skipObstructed=Skipped obstructed anchor.
+enderio.gui.travelAccessable.skipPrivate=Skipped private anchor.
 
 enderio.gui.telepad.configure.travel=Travel Settings
 enderio.gui.telepad.configure.telepad=Telepad Settings
--- a/src/main/java/crazypants/enderio/config/Config.java
+++ b/src/main/java/crazypants/enderio/config/Config.java
@@ -123,6 +123,9 @@ public final class Config {
 
   public static boolean travelAnchorEnabled = true;
   public static int travelAnchorMaxDistance = 48;
+  public static int travelAnchorCooldown = 0;
+  public static boolean travelAnchorSneak = true;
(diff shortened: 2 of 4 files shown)


tterrag1098:
This should probably have a config. It is client-sided after all.

Dwargkiller:
Please be patient while I sort out the mess I made.

Dwargkiller:
Pull request is finalized. Features outlined below:

1) Elevator functionality. If there is no valid selected target, it tries to select one above (if jump is used) or below (if sneak is used) within range. 
2) Adds sneak as an activation for targeted teleportation, with config option to disable it (default true)
3) Adds a configurable cooldown for travel anchors (default 0 ticks), though without any graphical representation of said cooldown.

Dwargkiller:
Did some work regarding access modes with vertical teleports:

1) If the possible target requires password and is NOT private (an unowned private anchor returns true on getRequiresPassword, owned returns false), open the password GUI. It does not do anything further, future attempts will open the gui if password was unsuccessful while teleporting if it was successful, as with other teleports.

2) Now skips private anchors it can't access as well as obstructed anchors.

TODO/Thoughts:

a) Add messages for skip events (Skipped obstructed anchor.. etc)
b) Make the password gui attempt teleportation on success and skip in vertical search if closed without success. Not sure how the password gui is handled and if it provides any hooks for open/close success/failure states.

eduardog3000:
Couldn't you change getRequiresPassword() from: 

```Java
@Override
  public boolean getRequiresPassword(EntityPlayer username) {
    return getAccessMode() != AccessMode.PUBLIC && !canUiBeAccessed(username) && !authorisedUsers.contains(PlayerUtil.getPlayerUUID(username.getGameProfile().getName()));
  }
```

to:

```Java
@Override
  public boolean getRequiresPassword(EntityPlayer username) {
    return getAccessMode() == AccessMode.PROTECTED && !canUiBeAccessed(username) && !authorisedUsers.contains(PlayerUtil.getPlayerUUID(username.getGameProfile().getName()));
  }
```

I don't see where a password would be required for anything but protected anyways.

Dwargkiller:
@eduardog3000 I'm planning to take a stab at reworking some of the ambiguity and edge cases in the anchor code, though I'll do that separate from this PR.

Dwargkiller:
Recent changes:

1) Locked travel anchors no longer provide gui. They are instead skipped in the search for targets when doing an elevator teleport.
2) A skipped anchor now produces a chat message.
3) Some cleanup of redundancy in functions (TileTravelAnchor)

eduardog3000:
I guess rebasing made our commit comments about OpenBlocks support disappear from this page, so I'm gonna link it here for future reference.

https://github.com/Lilylicious/EnderIO/commit/246eff7f7cf5afe13144d546eaf57e39dd723649#commitcomment-10817064

PR #2312, opened 2015-04-13
Update ru_RU.lang

Adaptivity:


resources/assets/enderio/lang/ru_RU.lang
```diff
@@ -915,7 +945,7 @@ enderio.gui.weather.requireditem=Требуется предмет: %s
 
 //Enchants
 enchantment.enderio.soulBound=Печать души
-description.enchantment.enderio.soulBound=Это зачарование оставляет предмет в инвентаре после смерти
+description.enchantment.enderio.soulBound=Предотвращает потерю предмета из инвентаря после смерти
```
tterrag1098: If this phrase could be condensed in any way it would be good, because this is now the line used on item tooltips under the enchantment entry, so this would take up quite a bit of space.

Adaptivity: Its already condensed as much as possible :)

Issue #2309, opened 2015-04-13
Reservoir Filling Issue


emperor06:
I have 2 issues with the Reservoir as multiblock:
1. Only 1 of the 4 blocks can be manually filled with a bucket. When clicking on any of the 3 others, nothing happens on first click then the next click will spill water in the world. Which block can be clicked to get filled depends on the order you placed the 4 blocks (and seems pretty erratic ^^).
2. Fluids cannot be piped in multiblock reservoirs.

Nota: single block reservoirs seems to work fine. It only affects the multiblock version.

6210classick:
what version are you experience this problem with ?

you fill the Reservoir with 2 buckets of water and it will keep generating water so why do you need to pipe water into it ???

is the multiblock formed correctly ? it needs to be 2x2x1

emperor06:
1. Using latest commit on branch master.

2. Commit f2f6bdd3 says: "Full IN+OUT: Tank, Reservoir, XP Obelisk, Soul Binder". Since "IN" is not working, I think it's better to mention it regardless of the reservoir common usage.

3. The multiblock is formed correctly (hard to do it wrong since the visual changes when it's complete). I have tried all 3 possible setups: 2x2x1, 2x1x2, 1x2x2. Same issue on all.

HenryLoenwind:
Darn, I know what happened. When I combined the right-click code into the super-class, I trashed the special coding the reservoir (as a multi-bock) needs. Oops.

emperor06:
:)

tterrag1098:
Fixed via https://github.com/SleepyTrousers/EnderIO/commit/215dcd425ce799a7f9b47ba968e55b535796a02d

PR #2314, opened 2015-04-13
Fix: Bucket Handling

HenryLoenwind:
* Restored Reservoir special multi-block handling (thx git)
* Restored sneaking behavior
* re #2309

### DIFF
--- a/src/main/java/crazypants/enderio/BlockEio.java
+++ b/src/main/java/crazypants/enderio/BlockEio.java
@@ -95,6 +95,10 @@ public abstract class BlockEio extends Block {
       }
     }
     
+    if (entityPlayer.isSneaking()) {
+      return false;
+    }
+
     if (te instanceof ITankAccess) {
       if (FluidUtil.fillInternalTankFromPlayerHandItem(world, x, y, z, entityPlayer, (ITankAccess) te)) {
         return true;
@@ -102,10 +106,6 @@ public abstract class BlockEio extends Block {
       if (FluidUtil.fillPlayerHandItemFromInternalTank(world, x, y, z, entityPlayer, (ITankAccess) te)) {
         return true;
       }
-    }
-
-    if(entityPlayer.isSneaking()) {
-      return false;
     }
 
     return openGui(world, x, y, z, entityPlayer, side);
--- a/src/main/java/crazypants/enderio/machine/reservoir/BlockReservoir.java
+++ b/src/main/java/crazypants/enderio/machine/reservoir/BlockReservoir.java
@@ -19,6 +19,7 @@ import crazypants.enderio.gui.IResourceTooltipProvider;
 import crazypants.enderio.machine.reservoir.TileReservoir.Pos;
 import crazypants.enderio.tool.ToolUtil;
 import crazypants.util.BlockCoord;
+import crazypants.util.FluidUtil;
 import crazypants.vecmath.Vector3d;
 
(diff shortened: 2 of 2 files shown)


tterrag1098:
I already fixed this, no?

HenryLoenwind:
Wasn't linked in...

Quick look says it might work, but leaves other parts of the code hanging. Sneaking you missed (only tested with buckets, I'd guess?).

tterrag1098:
I don't see why sneaking should prevent bucket filling. Did it before?

HenryLoenwind:
On some machines it dis, on some it didn't. That was part of the reason to consolidate the code.

Sneaking is the universal "right-click inverter". With all tanks I checked this means it allows source blocks to be picked up or placed into the world. Vanilla implements this in the bucket's code, all other mod's liquid containers leave it to the block.

src/main/java/crazypants/enderio/BlockEio.java
```diff
@@ -95,6 +95,10 @@ public boolean onBlockActivated(World world, int x, int y, int z, EntityPlayer e
       }
     }
     
+    if (entityPlayer.isSneaking()) {
```
tterrag1098: I think this check should be integrated into the next if clause (i.e. `if (!entityPlayer.isSneaking() && te instanceof ITankAccess)`) so that it does not interfere with later code that might want to function if the player is sneaking.

src/main/java/crazypants/enderio/machine/reservoir/BlockReservoir.java
```diff
@@ -69,11 +70,18 @@ public boolean onBlockActivated(World world, int x, int y, int z, EntityPlayer e
         }
         return true;
       }
+      if (FluidUtil.fillInternalTankFromPlayerHandItem(world, x, y, z, entityPlayer, tank)) {
```
tterrag1098: Why is this necessary?

HenryLoenwind: because the te *here* is the multi-block controller.

tterrag1098: But there is already code for that, see getInputTank and getOutputTanks in TileReservoir

src/main/java/crazypants/enderio/BlockEio.java
HenryLoenwind: I did not want to write

    if (!entityPlayer.isSneaking()) {
        code(1)
    }
    if (entityPlayer.isSneaking()) {
        code(2);
    }
    code3();

and I didn't want to put the remainder of the method into an "else" bacause that would change indent-level and I know, you don't like that in commits.

src/main/java/crazypants/enderio/machine/reservoir/BlockReservoir.java
HenryLoenwind: yes, as I wrote, you did that in a way I did not see.

And btw, your code leaves setDirty() hanging in the air now.

tterrag1098: How so?

HenryLoenwind: it does not contain the getController() code. So it can only fix the internal tanks if the player clicks the controller block by chance. And, because you present the tanks in the wrong order, in the other 3 cases, no water regens.

Issue #2313, opened 2015-04-13
Electromagnet keybinding in baubles slot.


hisorange:
Aye when you put the electromagnet in the baubles slot the keybinding not working as on/off switch anymore.

PR #2315, opened 2015-04-13
Fix: Magnet keyBinding undoing its own work

HenryLoenwind:
* Hotbar: Not harmful, but would be replaced by the server anyway
* Baubles: Harmful, as Baubles seems to syncs from client to server
* re #2313


tterrag1098:
How is that code "undoing its own work" ? Typically you set the client data and then send the info to the server for sync. Is that not what's occuring?

HenryLoenwind:
what's happening is:

code disables magnet in baubles slot
Baulbes syncs that to server
code tells server to desable magnet
server doesn't find active magnet the client is talking about and declares the client insane

And in the hotbar case, the local deactivation is not needed, as the server directly responds with an inventory update to the magnet-toggle message.

PR #2316, opened 2015-04-14
Tweaks for Farming Station

HenryLoenwind:
* Some code cleanup
* Harvest patches of plants in one go so it looks faster


HenryLoenwind:
Cleaning up in preparation for the structural changes needed to support Agricraft.

tterrag1098:
Why harvest patches at once? Won't that just result in it typically draining large chunks of power? Seems unnecessary.

HenryLoenwind:
I'll include the important bits in the rewrite that's needed for Agricraft support

PR #2325, opened 2015-04-15
New: Tanks can be used in item form

HenryLoenwind:


tterrag1098:
I think I will find a different way to do this eventually. #2574

PR #2329, opened 2015-04-16
New: Config value to disable upticking

HenryLoenwind:
* Config value to have machines reject an update rate that is higher
than the global tick rate.
* Default is unchanged (allow).

### DIFF
--- a/src/main/java/crazypants/enderio/TileEntityEio.java
+++ b/src/main/java/crazypants/enderio/TileEntityEio.java
@@ -5,6 +5,7 @@ import net.minecraft.network.NetworkManager;
 import net.minecraft.network.Packet;
 import net.minecraft.network.play.server.S35PacketUpdateTileEntity;
 import net.minecraft.tileentity.TileEntity;
+import crazypants.enderio.config.Config;
 import crazypants.enderio.machine.PacketProgress;
 import crazypants.enderio.network.PacketHandler;
 import crazypants.util.BlockCoord;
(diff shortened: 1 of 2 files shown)


src/main/java/crazypants/enderio/TileEntityEio.java
```diff
@@ -32,14 +33,19 @@ protected boolean shouldUpdate() {
...
+
   @Override
   public final void updateEntity() {
-    doUpdate();
-    if(isProgressTile && !worldObj.isRemote) {
-      int curScaled = getProgressScaled(16);
-      if(++ticksSinceLastProgressUpdate >= getProgressUpdateFreq() || curScaled != lastProgressScaled) {
-        sendTaskProgressPacket();
-        lastProgressScaled = curScaled;
+    if (Config.allowExternalTickSpeedup || worldObj == null || worldObj.getTotalWorldTime() != lastUpdate) {
```
Vexatos: OR if `worldObj == null`? This means this is getting called if the world object is null.....

MatthiasMann: worldObj can't be null in the tick method

HenryLoenwind: @Vexatos It means that it will not execute the method call on a null value but continue.

@MatthiasMann About half the doUpdate()s start with a null check. Hard to not type it after having seen it that often...

I'll clean them up later https://github.com/HenryLoenwind/EnderIO/issues/20

Issue #2333, opened 2015-04-18
Crash in Pathfinder pack (1.03) when configuring a face on stirling generator


ArbitraryHubris:
I was using the internal 3D gui to config a face and got this crash.  Now it crashes every time I start my world.  

http://pastebin.com/J3hkhfQ1

MatthiasMann:
I think that has been fixed - update EnderIO to 2.2.8 build 367 (available on Curse)

ArbitraryHubris:
I just tried EnderIO-1.7.10-2.2.8.367.jar and I still have the crash.  new log attached.

http://pastebin.com/SgLM5tQX

ArbitraryHubris:
I don't know why it's trying to render fused quartz.  I don't have any of those blocks sitting around.

HenryLoenwind:
Quite clear glass then?

Looks like something corrupted your world a little bit. Now there's a block with an invalid meta value, and that block just cannot take it. No idea where the corruption comes from, but the crash can at this location can be stopped.

ArbitraryHubris:
I've been thinking about what happened before the crash.  

I have a stirling generator placed directly next to FluxCrystals powered soil.  On the other side of the generator is a leadstone energy cell that has leadstone fluxducts going under the generator to power the soil.  Then, I picked up the stirling generator and dropped it into my Project-E alchemical tablet so I could duplicate it.  I created 3 of them from the tablet and placed them down to power the energy cell.  They all have octadic capacitors (removed when I duplicated them) and are burning aeternalis fuel from Project-E.  I was using the 3d user interface of the generator that sits between the energy cell and one of the powered soils.  I was trying to disable the side pointing at the powered soil when the crash occured.  Now just loading the world causes the crash to repeat.

I think that the fused quartz is a red herring.  Though, I did cook some fused quartz in an alloy smelter so there could be one or more left in that machine but it shouldn't have to be rendered.

Would you like a zipped copy of the world file?

ArbitraryHubris:
Another update: I didn't realize that there were two files generated per crash.  Now it looks clear that it's an interaction between enderio and fluxed crystals.

http://pastebin.com/8fQKitvG

ArbitraryHubris:
I'll open the issue on the fluxed crystals github if there is one.  Should I link that bug here or just close this one?

HenryLoenwind:
[...]

PR #2334, opened 2015-04-18
Clamp meta in glass renderer

HenryLoenwind:
* Don't crash on invalid block meta value
* re #2333


tterrag1098:
It's definitely something I've been meaning to do, now doubling the calls per-render is definitely not the best.

PR #2338, opened 2015-04-19
Cleanup: null checks in doUpdate()

HenryLoenwind:
They are not needed, the code calling doUpdate() would NPE if worldObj
was ever null at this point.


tterrag1098:
@HenryLoenwind You really should start using the EIO formatter settings, spaces before if parens is not proper format.

eduardog3000:
Spaces after `if` isn't proper, spaces after `for` is just fine though. According to the formatter settings, anyways.

tterrag1098:
Just as a sanity check, could you add an if (worldObj == null) return; statement in TileEntityEIO's updateEntity() ?

tterrag1098:
Actually nevermind, it's not possible

PR #2377, opened 2015-04-26
Fixed warnings

HenryLoenwind:
* Added missing @Overrides
* Removed unused fields and variables
* Removed impossible null checks
* Added missing null checks
* Fixed a couple of tiny bugs that were pointed out by the null checks warnings
* Added a whole bunch of @SuppressWarnings("unused") to parameters


tterrag1098:
On hold until after the EnderCore merge, currently it would affect too many things.

PR #2382, opened 2015-04-27
optimize and cleanup lights

MatthiasMann:


src/main/java/crazypants/enderio/machine/wireless/WirelessChargedLocation.java
```diff
@@ -24,13 +26,26 @@ private void updateChargers() {
...
     chargers.clear();
     lastChangeCount = wcc.getChangeCount();
-    wcc.getChargers(te.getWorldObj(), new BlockCoord(te), chargers);
+    final BlockCoord bc = new BlockCoord(te);
+    wcc.getChargers(te.getWorldObj(), bc, chargers);
+    Collections.sort(chargers, new Comparator<IWirelessCharger>() {
+      @Override
+      public int compare(IWirelessCharger o1, IWirelessCharger o2) {
+        int dist1 = o1.getLocation().distance(bc);
+        int dist2 = o2.getLocation().distance(bc);
```
tterrag1098: Because the actual distance is irrelevant, this should use distanceSq to save performance.

PR #2384, opened 2015-04-27
New: Cobbleworks

HenryLoenwind:
* New machine: Cobbleworks
* Produces cobblestone from energy.
* Can be upgraded with up to 3 of Crafter, Alloy Smelter, SAG Mill
* Upgrades are chained in order. They cannot be configured.
* Upgrades always produce all their outputs. For the crafter, only 3x3, 2x2, 1x1 and 3x2 recipes are checked.
* There are no material inputs. It will only work on cobble and the items produced from it. (Technically the upgrade slots are inputs.)
* Config: RF per cobble, RF discount for all 3 types of upgrades, RF discount for capacitor upgrades.

* Also in: Fixes to TechneUtil and some reworking of the rendering chain to allow easier access to the model.
* Telepad rendering should be checked, I'm not 100% sure I didn't break anything.

* Original file for the texture is included because it has some layers.


tterrag1098:
A few things noticed:

1. The cobbleworks is insanely fast, way too fast, even if it uses the power it should have a delay on it.

2. As a follow up to (1), the arrows should be animated

3. Behavior is odd, placing a sag mill in the first slow results in it creating a full stack of gravel before making any sand. Having any quite clear glass prevents a configured alloy smelter from making regular glass. etc.

HenryLoenwind:
1.) Got an idea on how to handle this. Let me try it out first.

2.) I fear the amount of network traffic on this. I kept the traffic it low on purpose, with only the usual content update and a bit to designate it was active the last 2 seconds. How about a "dummy" animation that simply runs on the arrows while it is active?

3.) I'm not sure I can quite follow you. Yes, it fills from the right on purpose, so it can refill a taken out stack in one operation (if the energy buffer allows).

HenryLoenwind:
1.) Speed now also is limited on max power output of the internal capacitor. Values may need tweaking.

2.) Animated arrows are in, they show which stages are working. Feel free to (have someone) make nicer animations.

3.) (we talked on irc about this)

HenryLoenwind:
2 commits because I want to branch off on the model and parts

tterrag1098:
I'm going to say no to this idea for now, though I do hate to discard so much work. If you can come up with something that makes it not just a machine that does what other machines can already do, we may revisit it.

PR #2389, opened 2015-04-28
Even more Baubles

HenryLoenwind:
* Baubles slots are now displayed in the Capacitor Bank. So the Magnet can be charged directly.


HenryLoenwind:
BTW: I checked the other paces where we access the Baubles inventory and found that they already correctly work around Baubles' bug by accident.

eduardog3000:
[Why not something like this?](https://gist.github.com/eduardog3000/6d1827dce94a4d3d8e92#file-capacitorbank-png)

HenryLoenwind:
Because I intended to look up how they render the images at home (armor-slot-like or hardcoded) and forgot in the mess with Baubles' bug. Feel free to PR this, you already did the work.

eduardog3000:
You can just rebase it into this branch.

PR #2412, opened 2015-05-02
added Farming Station support for B0bGary's Growable Ores

lehjr:
This was tested and works well. I don't like the wall of text, but I'm not seeing a better way of doing this.


lehjr:
Good suggestion. I'll try and hammer that out this weekend.

tterrag1098:
I used your code as a nice hint, but I found a shorter way to do it. Thank you!

PR #2418, opened 2015-05-03
TConstruct support for alloy smelter

Adaptivity:
1) Added common/uncommon blocks to SAG Mill, formatting fixes, updated TE/TF support.
2) Added TConstruct alloys to alloy smelter, formatting fixes. Closes #768
About Pig Iron: Currently Pig Iron recipe in Tcon is messed up (will be fixed in next version), recipe in Ender IO gives two ingots instead of 1 in Tcon since Coagulated Blood gives 160 mb of blood but 1 ingot requires only 80. So, I decided to make recipe require double amount of materials and gives 2 ingots instead of 1.
Suggestions for power consumption are welcome.


tterrag1098:
I don't think I'm going to do this. I already have enough of a problem with the CoFH team because of the Enderium recipe. This might be a job for @Drullkus in ThermalSmeltery.

PR #2426, opened 2015-05-06
New: Dark Hoe

HenryLoenwind:
Just a basic dark steel hoe. Empowered will do AOE.


tterrag1098:
Fork upgrade will be used instead of this.

hron84:
@tterrag1098 will it be used for Farming Stations? Basically, I need a chargeable hoe to be used in Farming Stations.

PR #2439, opened 2015-05-07
New: Fork Upgrade

HenryLoenwind:
Changed tool upgrade system to be more flexible. Moved "place block" into
an upgrade. Added hoe upgrade.

Note: Upgrades that use the "right click" slot are mutable exclusive. Same
for "shift right click", but there's only one upgrade atm.


eduardog3000:
How about a way to cycle through which upgrade is active? Y cycles through Yeta Wrench settings right? Maybe Y for right click upgrades and shift-Y for shift right click upgrades?

MatthiasMann:
Good idea - but you can use the same key - as it doesn't make sense to cycle through things for an item not in your hand.

eduardog3000:
Yeah, the Yeta Wrench already uses Y, so the other tools can use it as well.

tterrag1098:
I had a different idea for the Fork upgrade (or whatever I was going to call it, I suck at names), so I'll put this on hold until after the EnderCore update.

HenryLoenwind:

Issue #2485, opened 2015-05-18
[Feature Request] Upgrade conduits in-place


dgbonomo:
When you have a lot of conduits built and wish to upgrade to a higher tier of the same type, it can get very challenging. I'm not sure how feasible it would be, but it might be nice if there were craftable conduit upgrades that you could right-click to upgrade an existing conduit in-place.  There may be other challenges with this (I didn't think it through much) but I just thought I'd at least get it out there.

HenryLoenwind:
It may be possible to change the code that rejects placing a conduit if there already is one of that type there to "just" exchange them. I'll have a look, as I've had the same problems in the past ;)

tterrag1098:
#2486 Was nonfunctional, this has been fixed via https://github.com/SleepyTrousers/EnderIO/commit/5a6a35ac144d286451ce33dd7a86d2b0f9a029e3 and is present in [build 402](http://ci.tterrag.com/job/EnderIO/402/).

HenryLoenwind:
Any post-mortem why it worked for me and not others?

And BTW: You sure your want to spawn the item entity in the center of the conduit block? That way it has the tendency to go everywhere, just not where the player can pick it up...

tterrag1098:
@HenryLoenwind in most circumstances, `placeAt` would not be null, thus it would always return `true` before ever getting to the replace code. Moving it to `onItemUseFirst` guarantees it to override other behavior.

PR #2486, opened 2015-05-18
Replace Conduits

HenryLoenwind:
* When trying to place a conduit where different one of the same type already exists, this will drop the existing conduit and place the new one. (Old: do nothing)
* re #2485

### DIFF
--- a/src/main/java/crazypants/enderio/conduit/AbstractItemConduit.java
+++ b/src/main/java/crazypants/enderio/conduit/AbstractItemConduit.java
@@ -5,6 +5,7 @@ import java.util.List;
 import net.minecraft.block.Block;
 import net.minecraft.client.renderer.texture.IIconRegister;
 import net.minecraft.creativetab.CreativeTabs;
+import net.minecraft.entity.item.EntityItem;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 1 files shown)


src/main/java/crazypants/enderio/conduit/AbstractItemConduit.java
```diff
@@ -103,6 +104,31 @@ public boolean onItemUse(ItemStack stack, EntityPlayer player, World world, int
...
           return true;
+        } else {
+          // There's not really any good way to compare conduits...
+          IConduit existingConduit = bundle.getConduit(getBaseConduitType());
+          ItemStack existingConduitAsItemStack = existingConduit.createItem();
+          if (!crazypants.util.ItemUtil.areStacksEqual(existingConduitAsItemStack, stack)) {
+            if (!world.isRemote) {
+              IConduit newConduit = createConduit(stack, player);
+              if (newConduit == null) {
+                System.out.println("AbstractItemConduit.onItemUse: Conduit null.");
```
TheJulianJES: Do you need that?

HenryLoenwind: The alternative would be to throw an exception. ;)

The only way that can happen is when there's a bad conduit type. That should be noticed while developing the conduit type, I'd say. This is not a "play time" condition.

And, btw, I just copied the existing code from just above.

TheJulianJES: No, I mean the System.out thing..

TheJulianJES: Okay :)

Issue #2568, opened 2015-06-10
Add more powerful energy conduit?


widd:
Been using EnderIO for a long time now and I think that some sort of high-cost/high-power cable is in order. Endgame content where massive amounts of power are required forces some poor design choices for complex machine setups.

Keeping in touch with the measure of conduit throughput, could there be some sort of "ultimate" energy conduit with around 80k rf/t throughput? Higher cost would obviously be required.

The limits of the dimensional transceiver make sense but seeing as capacitor banks have cumulative throughput power dependent on their multiblock size, I feel as if this should've been in the mod awhile ago.

MatthiasMann:
You can just use multiple connects of the ender energy conduit - as the 20k RF/t limit is per connection.

widd:
I'm well aware - that's why I suggested around 80k rf/t. Realistically I cannot use every single side of a machine as most machines have input/output on at least one side, sometimes two or more. Most EnderIO machines do not have this issue, but capacitor banks do. Splitting a ~3  billion total RF capacitor bank between 20k rf/t conduits is not very practical.

HenryLoenwind:
20k is more than enough if Ender Io is used on its own. If you use it in combination with other mods, you should configure the mods to work nicely with each other. One of those config options is the RF limit of our conduits and cap banks.

PR #2572, opened 2015-06-11
4th tier energy conduit - 'Ultimate'

widd:
For #2568


MatthiasMann:
EnderIO has only 3 tiers of energy conduit for a reason. Adding a 4th will not add to the game play at all. And if people are playing a pack with such greatly unbalanced power values they can adjust the config values for the existing tiers.

PR #2664, opened 2015-07-21
Change glass pane recipe to match the Vanilla one

Quantum64:


tterrag1098:
This has the same problem just in reverse. The vanilla recipe you can put in either the top 2 or bottom 2 rows.

PR #2739, opened 2015-08-25
SAG Mill support for blocks

Adaptivity:
Reworked version of #2418 but now without touching CoFH/Tcon stuff. Only added blocks support for existing stuff.

### DIFF
--- a/resources/assets/enderio/config/SAGMillRecipes_Core.xml
+++ b/resources/assets/enderio/config/SAGMillRecipes_Core.xml
@@ -445,7 +445,7 @@
     </recipe>
   </recipeGroup>
 
-  <recipeGroup name="Common Ores/Ingots" >
+  <recipeGroup name="Common Ores/Ingots/Blocks" >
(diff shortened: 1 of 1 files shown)


Adaptivity:
@tterrag1098 also, why it's a problem to add SAG mill support for CoFH ingots/blocks? This even not a balance change. I know that alloy creating might be unbalanced but this is just processing of 1 ingot for 1 dust and 1 block for 9 dusts. This just can't be unbalanced. It's just simple recipes. Maybe it can be reconsidered?

tterrag1098:
Who said I had a problem with CoFH ingots/blocks? AFAIK we already support all of them. The problem I had was with tcon alloys.

Adaptivity:
Ah, I misread your comment on #2418 :) Can I then add [these recipes](https://github.com/SleepyTrousers/EnderIO/pull/2418/files#diff-c7a1baa0b5a92796f4288d4125f10fe4R733)?

tterrag1098:
Yeah

PR #2742, opened 2015-08-26
Ore Dictionary support for recipes

Adaptivity:
Added oredict support where possible. I'm not familiar with Ender IO oredict registration but it should work. I can't test this because Ender IO won't build since this commit https://github.com/SleepyTrousers/EnderIO/commit/576199774f228599913406fd318d09ce86fbd08f
I'll fix formatting if there are no other problems. Chests and crafting tables were registered to support Wood Stuff mod by ganymedes01.

### DIFF
--- a/src/main/java/crazypants/enderio/material/MaterialRecipes.java
+++ b/src/main/java/crazypants/enderio/material/MaterialRecipes.java
@@ -40,7 +40,7 @@ public class MaterialRecipes {
   }
 
   public static void registerOresInDictionary() {
-    //Ore Dictionary Registeration
+    //Ore Dictionary Registration
     OreDictionary.registerOre("dustCoal", new ItemStack(EnderIO.itemPowderIngot, 1, PowderIngot.POWDER_COAL.ordinal()));
     OreDictionary.registerOre("dustIron", new ItemStack(EnderIO.itemPowderIngot, 1, PowderIngot.POWDER_IRON.ordinal()));
     OreDictionary.registerOre("dustGold", new ItemStack(EnderIO.itemPowderIngot, 1, PowderIngot.POWDER_GOLD.ordinal()));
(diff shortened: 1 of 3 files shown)


src/main/java/crazypants/enderio/material/MaterialRecipes.java
```diff
@@ -60,6 +60,18 @@ public static void registerOresInDictionary() {
     OreDictionary.registerOre("sand", new ItemStack(Blocks.sand, 1, OreDictionary.WILDCARD_VALUE));
     OreDictionary.registerOre("ingotIron", Items.iron_ingot);
     OreDictionary.registerOre("ingotGold", Items.gold_ingot);
+	OreDictionary.registerOre("blockIron", Blocks.iron_block);
```
tterrag1098: This is not where we do standard oredict registrations, and most of these are already done by forge. See [OreDictionary.java](https://github.com/MinecraftForge/MinecraftForge/blob/1.7.10/src/main/java/net/minecraftforge/oredict/OreDictionary.java#L58-L128).

If you have any extras to add do it [here](https://github.com/SleepyTrousers/EnderCore/blob/master/src/main/java/com/enderio/core/common/OreDict.java),

Adaptivity: I know that these in Forge. I was wondering why ingotIron, ingotGold and etc registered here, this is why I've added new ones. Should I remove those that I've added? And just add craftingTableWood and chestWood to [OreDict.java](https://github.com/SleepyTrousers/EnderCore/blob/master/src/main/java/com/enderio/core/common/OreDict.java)?

tterrag1098: Yes

Adaptivity:
Should be done. Hope you can merge two commits since Github for Windows doesn't allow to squash...

tterrag1098:
You can still squash with the client, you just need to open up a shell. 

`git rebase -i HEAD~2`
change the first commit from pick to squash
save file
`git push -f origin oredictsupport`

tterrag1098:
Nope you made it worse :P

tterrag1098:
http://davidwalsh.name/squash-commits-git

Adaptivity:
I know :D Commit squashing never worked for me in Github for Windows. I'll try to figure this out.

Adaptivity:
Better create new PR :)

PR #2745, opened 2015-08-27
Ore Dictionary support for recipes

Adaptivity:
#2742


tterrag1098:
Thanks, I thought I caught all of these but apparently not.

Issue #1712, opened 2015-01-18
[Suggestion] More TE support


Karmacopper:
Please add CofH Holding enchant to the Enchanter. Maybe use leather as the catalyst.

tterrag1098:
This could easily be a PR instead of an issue >.>

Issue #880, opened 2014-08-16
Soul Stealer for ender io enchantment table  


tgur32:
this is what i have for the enchament i couldnt get diamonds to go in or these heads was putting it in the EnchanterRecipes_User file  trying to get it to be 10 zombie heads per level of soul stealer i took out the </ because it wasnt showing up

enchaterRecipes>
enchantment name="Soul Stealer" costPerLevel="10">
    itemStack modID="minecraft" itemName="zombie heads" /> 
  enchantment>
enchaterRecipes>

p75369:
Here we go, the config for the Soul Stealer enchant using a Soul Shard.

Not so sure about the cost. It does have 5 levels, so you'd need to make 16 books to max it.

And I don't know what's up with the "item.null", but that's what NEI reporting and it works so...

<enchantment name="enchantment.Soul Stealer" costPerLevel="2">
    <itemStack modID="SSR" itemName="item.null"/>
  </enchantment>

tterrag1098:
item.null is not the item's registered name. Use NEI's item dump to get the real name.

p75369:
I did, that's where I got it from, full line from the dump:

SSR:item.null,6750,false,SSR,ssr.gameObjs.items.SoulShard

tterrag1098:
Please post your entire file and if you can, upload the NEI dump so I can confirm.

p75369:
Think this'll work, not used to sharing files.

https://drive.google.com/file/d/0B4UCt4Fls0GUR1QzNGhrcE5UNmM/view?usp=sharing

PR #2774, opened 2015-09-06
#1712 #880 More enchants for enchanter

Adaptivity:
Soul stealer enchantment should be ok
CoFH enchantment catalysts doesn't really makes sense but... they are good for me. If anyone can suggest better catalysts I can change them. But keep in mind that they are should be from TE/TF/TD mods.
Holding uses Compressed Sawdust
Multishot uses 2 Slags

### DIFF
--- a/resources/assets/enderio/config/EnchanterRecipes_Core.xml
+++ b/resources/assets/enderio/config/EnchanterRecipes_Core.xml
@@ -105,5 +105,14 @@
   <enchantment name="enchantment.repair" costPerLevel="6" >
     <itemStack modID="minecraft" itemName="anvil" />
   </enchantment>
+  <enchantment name="enchantment.soul_stealer" costPerLevel="6" >
+    <itemStack modID="SSTOW" itemName="sstow_soul_shard" />
+  </enchantment>
+  <enchantment name="enchant.cofh.holding" costPerLevel="6" >
+    <itemStack modID="ThermalExpansion" itemName="material" itemMeta="513" />
+  </enchantment>
+  <enchantment name="enchant.cofh.multishot" costPerLevel="6" >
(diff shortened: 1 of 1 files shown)


tterrag1098:
What even are those items? Why not use oredict?

Adaptivity:
Soul stealer -> Soul shard
Holding -> Compressed Sawdust
Multishot -> 2 Slags
Are they oredicted? They are not oredicted in TE.

Adaptivity:
So... what do you think?

Yulife:
Is it done the same way in TE or what does multishots have to do with a byproduct of an induction furnace o_o

Adaptivity:
Is TE have catalysts for their enchants? I don't think so. The main problem is that catalysts should be from CoFH mods (They can be from vanilla too but CrazyPants recommended to use them from mods). And I don't know which items will fit better. If you can suggest better catalysts, go ahead, I'll change them.

Yulife:
Multishot: Aerotheum Dust, Oredict is "dustAerotheum"
Holding: Reinforced Strongbox, Modname: ThermalExpansion; Item Name: Strongbox; Metadata: 3

Adaptivity:
@tterrag1098 should say final word.

tterrag1098:
How is aerotheum made?

cloakable:
One sandstone, one blitz powder (Thermal Foundation mob), one redstone, one niter gives two aerotheum dust.

tterrag1098:
Aerotheum seems reasonable for multishot, and I also like the strongbox for holding, though it seems a bit redundant (since it only goes on strongboxes and satchels) :P

Adaptivity:
So, should I update this?

tterrag1098:
Yes

Issue #2743, opened 2015-08-27
enderio glider upgrade kicks me from my server.


ziba1:
Hi, me and a friend have our own server running the 1.7.10 pack which has the enderIO mod.

We just updated recently the server recently to the 0.7.9f version which has the 2.3.0.393 version of the mod installed

Now randomly whenever my friend is using his enderio glider upgrade, I am  booted from the server, not necessarily instantly, but within like 30 seconds or so of him using it.  There is nothing specific I am doing when this happens, I can just be stood still.

This is the console message I get when it happens:

[23:18:10] [Server thread/INFO]: zibafu lost connection: TranslatableComponent{key='disconnect.genericReason', args=[Internal Exception: java.io.IOException: Une connexion Ã©tablie a Ã©tÃ© abandonnÃ©e par un logiciel de votre ordinateur hÃ´te], siblings=[], style=Style{hasParent=false, color=null, bold=null, italic=null, underlined=null, obfuscated=null, clickEvent=null, hoverEvent=null}}

Hes stopped using the glider since then, and Ive stopped getting kicked.

HenryLoenwind:
Could you please also have a look in the client log files if there are more error messages? The above is just a "lost connection" and not really helpful for finding the cause...

ziba1:
will do, probably tonight, will get my mate to use his glider again and try to recreate the problem.

Which sods law will say that I then don't disconnect haha

ziba1:
Ok, literaterly just logged on with my mate, asked him to put his suit on and deploy the glider, soon as he did I was kicked, so it definitely seems to be something to do with the glider

Anyhow, here are the logs

http://paste.ubuntu.com/12208845/

http://paste.ubuntu.com/12208862/




Cheers guys.

tterrag1098:
Quite a few errors in those logs aside from you getting kicked. But still nothing helpful.

ziba1:
You think it could be a combination of things ?

I might update the modpack on my server today, then get my mate to test it again later

rladams:
Using Revolution pack 1.0.10.0, EIO 2.3.0.407_beta on a server. All players get kicked when any ability (jump in this case) is toggled by the player using the dark steel armor.

tterrag1098:
Is your world in offline mode? Also this is unrelated.

rladams:
No, we run with online mode and full auth.

[...]

PR #2780, opened 2015-09-07
Don't crash if in MP a client doesn't have another player's profile loaded

HenryLoenwind:
re #2743


HenryLoenwind:
This may not be the root cause, but null checking seldom hurts...

MrNuggelz:
The Execption seems to come from the EntityPlaye being null. I altered your code, to check if the player is null:
GameProfile gameProfile = player == null ? null : player.getGameProfile();
Since then the exception didn't occur.

HenryLoenwind:
Thx for testing, I adapted the PR to accommodate that case.

tterrag1098:
Have you tested this on a server to be sure that the controls still function?

MrNuggelz:
I have it running on my MP server and i haven't found a problem. Controlls work fine.

PR #2792, opened 2015-09-08
small enhancements(?) related to TinkersConstruct and Mekanism

p-schneider:
TileFarmStation: Allow almost completely broken Tools from Tinkers Construct to be extracted from the Farming Station.
With this modification I'm able to automate the use of TinkersConstruct Tools in the Farming Station so it is easy to automatically recharge a Tool with the flux modifier. 

TileKillerJoe: The Killer Joe now also accepts Tinkers Construct Weapons (those don't extend ItemSword)
The effects (modifiers) of the Weapon do apply, so the attack works as expected, but the weapon does not render. The Killer Joe looks like it has no weapon equipped.
Maybe allowing the Weapon to be extracted like I did it with the Farming Station would be useful, too. I haven't used the Killer Joe much, and I've not yet tried to automate it.

TileEntitySolarPanel: The Mekanism Solar Evaporation Plant (aka Salination Plant) can now be "powered" by EnderIO Photovoltaic Cells. (seesSun() uses the same constraints Mekanism uses with their Solar Panel.)
I like to play with Mekanism but I only use the base package which does not include the Generators. I looked for another way to run a Solar Evaporation Plant so I decided to implement the required Interface on the Photovoltaic Cell form EnderIO. I did not restrict it to the Advanced Photovoltaic Cell, the (basic / orange) Photovoltaic Cell does also work.

Both of the TinkersConstruct related modifications work without any additional API requirements.
Since the mekanism API is already used, the Photovoltaic Cell modification doesn't need an additional API either.
I have used this modifications for quite some time on my singleplayer worlds and I think it's time to at least share the idea and the code I wrote. Feel free to cherry pick whatever you like if you think it would be useful.

### DIFF
--- a/src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
+++ b/src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
@@ -294,6 +294,23 @@ public class TileFarmStation extends AbstractPoweredTaskEntity {
         EnchantmentHelper.getEnchantmentLevel(Enchantment.fortune.effectId, stack));
   }
 
+  @Override
+  public boolean canExtractItem(int slot, ItemStack itemstack, int side) {
+    if(super.canExtractItem(slot, itemstack, side)) {
+      return true;
+    }
+    //allow almost completely broken tools from Tinkers Construct to be extracted to get them repaired/recharged
+    if(slot >= minToolSlot && slot <= maxToolSlot && itemstack.getItemDamage() >= 95) {
+      Class<?> c = itemstack.getItem().getClass();
+      do {
+        if(c.getName().equals("tconstruct.library.tools.HarvestTool"))
+          return true;
+        c = c.getSuperclass();
+      } while(c != null);
+    }
+    return false;
+  }
+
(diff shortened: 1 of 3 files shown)


tterrag1098:
I don't like the way any of these are done.

The Tcon compat is hacky and expensive, and the `seesSun()` check is broken with non-desert biomes which may still not have rain.

tterrag1098:
Closed due to unresponsiveness

p-schneider:
I have to admit that my modifications related to Tcon are hacky and expensive and I see why you don't like to have hacky and expensive code in EnderIO.
However the `seesSun()` is exactly the same logic Mekanism uses with their solar panel. (Unless I messed up summarizing the boolean expression.) If there is a non-desert biome without rain and it rains in the world the salination plant would just stop producing until the rain ends just like it would do with the original mekanism solar panel equipped.

Eventually I don't mind it not being merged/accepted because I can always compile it myself if I like to use my modifications in my game.

tterrag1098:
Then mekanism has a bug, and I don't want to transfer that bug over to our codebase with copypasta code.

PR #2827, opened 2015-09-19
Use seperate texture for machine bottom

HenryLoenwind:
The side looks a bit silly on the bottom, as it has an up and a down. "machineTemplate.png"/"blockSoulMachineBlank.png" are nice and
symetrical. Should they be copied over to
"machineBottom.png"/"blockSoulMachineBottom.png"?


tterrag1098:
I think that's a decision for @CyanideX :P

CyanideX:
Sure, why not? :stuck_out_tongue_winking_eye:

Issue #2817, opened 2015-09-15
EE3 Support


jztech101:
EE3 support for its items?

HenryLoenwind:
I've been looking into it yesterday. Most if Ender IO's items would have EMC values, but there are bugs in EE3 that prevent that. Main one is that is fails to compute EMC as soon as an item in the recipe has multiple oreDict names. I plan on adding the support, but until EE3 gets an update that addresses this, it's not high on my TODO list---I don't want to add code to compute EMC for items that EE3 should compute itself.

jztech101:
Hmm I'm confused. Because EE3 seems to lack EMC for most of the alloys and stuff. I assumed it ws because they go through a machine that EnderIO needed to inform EE3 about. Should those be computed automatically?

HenryLoenwind:
No, alloys and liquids need to be added from our side. But EE3 doesn't manage to compute EMC for e.g. the Basic Gear---which is just 4 cobble and 4 sticks. Depending on the installed mods, many or even most of the basic recipes---which only contain vanilla items---fail to compute. That is what makes it not so useful to add the alloy and fluid emc values.

jztech101:
Hmm. Then how are Furnace recipes auto-computed? Can't you simply add the
Alloy Furnace as another "furnace" for alloys? Also, as for the failure to
calculate EE3 I have found that installing EE3helper resolves the issue.
Though that would explain why I have to Forge Lexicon one mods Copper Ingot
to another to get the proper EMC when the mod claims to add EMC values for
all the equivalent ore dict items.

tterrag1098:
The fact that you need *another* mod to make EE3 work properly makes me less inclined to add support for it.

jztech101:
The thing is, it seems the development version of EE3 appears to fix the issue (based on what I have seen on Soaryns Stream and Dire's series) with the lack of DynEMC values being calculated. the problem with that is.. it seems that pahimar has yet to do a proper release. And this version just does not work properly. I have heard rumors of buildcraft being an issue among other things.

HenryLoenwind:
As tt said. However, I plan to do so. It is not much hassle to register our liquids and alloys. *If* the automatic calculation for crafting recipes works.

It'd be nice if EE3 had an IMC API. It's much simpler to send a couple of messages than to include yet another batch of code.

HenryLoenwind:
[...]

PR #2831, opened 2015-09-20
EE3 EMC values

HenryLoenwind:
* Added EE3 EMC values
* Fixed reversed ender pearl dust registration logic
* re #2817

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,6 +64,7 @@ dependencies {
     compile "com.mod-buildcraft:buildcraft:${buildcraft_version}:api"
     compile "mcp.mobius.waila:Waila:${waila_version}_${minecraft_version}:dev"
     compile "net.sengir.forestry:${forestry_version}:api"
+    compile "com.pahimar.ee3:EquivalentExchange3:${ee3_version}:dev"
 
     // WAILA ships really old CCC/CCL versions, so we force the new versions
     compile "codechicken:CodeChickenLib:${minecraft_version}-${ccl_version}:dev"
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,6 +13,7 @@ ccc_version=1.0.7.46
 ccl_version=1.1.3.138
 ae2_version=rv1-stable-1
 endercore_version=1.7.10-0.1.0.24_beta
+ee3_version=1.7.10-0.3.507
 forestry_version=forestry_1.7.10:3.6.2.19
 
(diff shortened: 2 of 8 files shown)


tterrag1098:
I refuse to believe this code is safe without EE3 installed...everything I know says it's not.

HenryLoenwind:
It is. It only contains calls to static methods, those are not inspected until they are actually executed. No field or local variables of some unloaded type, not interfaces, no inheritance. Basically the same thing as calls to a @sideonly-removed method.

    INVOKESTATIC com/pahimar/ee3/api/exchange/EnergyValueRegistryProxy.addPreAssignedEnergyValue (Ljava/lang/Object;F)V

tterrag1098:
But you are executing them without checking if EE3 is loaded. the `if (IS_EE3_LOADED)` check is sufficient??

HenryLoenwind:
as it is

    public static final boolean IS_EE3_LOADED = Loader.isModLoaded(EE3_ID);

it should be. I could throw in a check for the API version, but that'd make us the only mod to do so.

And yes, I have studied about every mod that uses the ee3 api and can be found by Google. Had to, to get an idea how to even use that thing.

tterrag1098:
What I'm saying is that it was my understanding that when a method is invoked, all references are resolved. Why is this not the case here?

HenryLoenwind:
Because there is no reference to any object there. Only an INVOKESTATIC that takes one parameter, the fully qualified name of the *static* method to invoke. Or to quote the spec: "after the invokestatic opcode is a 16-bit unsigned integer index. This is the index of an entry in the constant pool." which contains a CONSTANT_Methodref entry which points to a CONSTANT_Class entry, all of which only contain references to CONSTANT_Utf8_info values. ( https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html ) So the actual class lookup/loading only happens when the INVOKESTATIC is actually executed and the Strings are resolved into class and field objects. Until that time, they are only IDs into the constant pool and Strings in it.

An INVOKEVIRTUAL (non-static method call) would be different. There an object to call the method on has to exist in the calling context, even if it isn't explicitly assigned to a variable. You can see that in the actual com.pahimar.ee3.api.exchange.EnergyValueRegistryProxy.addPreAssignedEnergyValue() method. There the call is dispatched to an object (API -> mod), but an additional layer of indirection must be used for this to work.

tterrag1098:
Ok, fine, I *guess* I believe you now :stuck_out_tongue:

PR #2834, opened 2015-09-21
Generic GUI classes handling

HenryLoenwind:
This moves the creation of Container and Gui classes for all machines up into AbstractMachineBlock.

Many of the removed handler were not type-safe. This solution is. It also saves much copy&paste code. The additional overhead is very small, it consists of a switch() statement and calling the constructor indirectly. As this only happens when a player opens a GUI, that won't matter. The "no GUI" case is a single null check.

Feel free to pull this up into EnderCore. I still have no combined workspace to develop combined EC/EIO changes. Sorry.


HenryLoenwind:

tterrag1098:
This eliminates some code but decreases maintainability. It's pretty ugly as well. I don't think I'm going to go with this method of doing things.

PR #2835, opened 2015-09-21
Use type-safe getTileEntity() in blocks

HenryLoenwind:
This builds upon #2834, which eliminates so many getTileEntity() calls. (Either pull only this or the other one first, in which case I'll rebase this one.)

Once https://github.com/SleepyTrousers/EnderCore/pull/17 is in Ender IO's required EnderCore, I'll do the calls that have a IBlockAccess.


HenryLoenwind:
I can rework this to be standalone, as you don't like the generic handling one.

HenryLoenwind:
For reference, another report: http://pastebin.com/nB0Lh1C0

PR #2842, opened 2015-09-22
Machine bottom now is its own texture

HenryLoenwind:
Also fixed obelisk rendering. It now uses the textures that are defined in the block and has a nice bottom texture. When rendered as items, they will now also have their own texture instead of using the same one for all of them.

Obelisk


HenryLoenwind:
PS: Yes, the top texture of the obelisks changed. Instead of using the side texture for all side, the renderer now uses the texture that is declared in the BlockObelisk classes. I and IRC think that looks better, but I'd have no problem changing the obelisks to return another texture for their top.

Issue #2828, opened 2015-09-19
Lapis Block to Lapis Lazuli Recipe Bug


andersleet:
I noticed that in NEI the recipe for Lapis Lazuli from a Lapis Block (not dense, even though in NEI both say they yield the same) using an alloy smelter (in furnace mode) says you will get 18 Lapis Lazuli. However, you only get one. Since I found this out on a server I play on with friends that has many, many mods loaded, I created an ender IO-only world and was able to recreate the same issue.

HenryLoenwind:
Please check what happens in a vanilla furnace and give your Ender IO version.

andersleet:
1 Lapis from vanilla furnace.

I am using Ender IO version 1.7.10-2.2.8.381

HenryLoenwind:
Ok, then Ender Io is working as expected---in furnace mode it mimics the vanilla furnace.

I'll have a look at the NEI display later.

PR #2845, opened 2015-09-23
Keep care not to store ItemStacks of furnace recipes

HenryLoenwind:
* re #2828 (may be a fix, or may be an unrelated mistake)

### DIFF
--- a/src/main/java/crazypants/enderio/material/OreDictionaryPreferences.java
+++ b/src/main/java/crazypants/enderio/material/OreDictionaryPreferences.java
@@ -62,7 +62,7 @@ public final class OreDictionaryPreferences {
     }
 
     if(result == null) {
-      result = stack;
+      result = stack.copy();
     }
     stackCache.put(key, result);
     return stack;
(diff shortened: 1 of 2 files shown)


src/main/java/crazypants/enderio/material/OreDictionaryPreferences.java
```diff
@@ -62,7 +62,7 @@ public ItemStack getPreferred(ItemStack stack) {
     }
 
     if(result == null) {
-      result = stack;
+      result = stack.copy();
     }
     stackCache.put(key, result);
     return stack;
```
tterrag1098: Should not maybe this return stack.copy()? Then we solve the problem at the root.

HenryLoenwind: I though about it, but it seems more prudent to always expect to get a "dirty" stack from any method---Forege and MC methods all do that.

HenryLoenwind:
I did a little thinking; what about this solution?...:

https://github.com/HenryLoenwind/EnderIO/commit/2db0901ddf68eba77da9439c12349f9c8c0c2637

That could be extended to all places where we store ItemStacks.

PR #2848, opened 2015-09-23
Various null checks

HenryLoenwind:
Some potential NPEs, some impossible null checks, some reordering to get
rid of null warnings, and even one or two logic errors

### DIFF
--- a/src/main/java/crazypants/enderio/block/BlockDarkSteelAnvil.java
+++ b/src/main/java/crazypants/enderio/block/BlockDarkSteelAnvil.java
@@ -71,6 +71,7 @@ public class BlockDarkSteelAnvil extends BlockAnvil implements IResourceTooltipP
     return true;
   }
 
+  @Override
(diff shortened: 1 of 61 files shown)


tterrag1098:
I don't like unnecessary null checks, something being null where it shouldn't be is a bug elsewhere.

HenryLoenwind:
With the exception of one or two "oh compiler shut up"-null checks, all new null checks are on paths where null is a valid value.

BTW: Isn't that a lovely pearl of 'what's my line?':

         if(cb != null) {
           totalStored += cb.doGetEnergyStored();
         }
         ItemStack[] inv = cb.inventory;

;)

tterrag1098:
Regardless I'd prefer if this was split into multiple commits or even PRs. "logic errors" doesn't describe what you changed.

HenryLoenwind:
ok, I'll divide it up. np

PR #2851, opened 2015-09-24
Cache some rendering calcs

HenryLoenwind:
I grabbed a random "TODO". Even though this calc is not done every
frame, this still saves some time.


tterrag1098:
Make sure to add it as a reload listener if you are going to cache UVs

`RenderUtil.registerReloadListener(...)`

HenryLoenwind:
That's texture reload? Thx, I thought about it early but the thought got lost on the way. And I didn't know how to.

I'll add it tonight.

If you think this way of caching is a good idea, I can pull out that caching interfaces into a public class so we can use it for more rendering.

tterrag1098:
Would be nice for the future. For now, this works.

PR #2866, opened 2015-09-30
Caching in HyperCube

HenryLoenwind:
Another random "TODO".

The cached operation is not very big, but the caching overhead should
save time for targets that have more than 1 or 2 input slots. I also
optimize object creation (delayed or even skipped ItemStack.copy() and
try to keep the cache's memory footprint down (36 entries max).


MatthiasMann:
You are aware that the hyper cube is the old/deprecated item which is only left for compatibility reasons?

tterrag1098:
Indeed

Issue #2867, opened 2015-10-01
[Suggestion] Spawning Error Handlin


agowa338:
As mentioned in IRC, can I suggest one feature? An error message in the GUI, if the hight is wrong. Eg. "Lower the spawner by one" or "no room to the sides to spawn"

or maybe like MatthiasM2 sayd: how about spawning an ender dragon to make more room for spawning :D

HenryLoenwind:
Start with a Wither. Only spawn the dragon if that doesn't help...

For the text I imagine the hovering text the farming station uses.

agowa338:
How would a wither fit into the "Ender*" theme?

PR #2868, opened 2015-10-02
Spawner: Notify with hover text on spawning failures

HenryLoenwind:
re #2867

### DIFF
--- a/resources/assets/enderio/lang/en_US.lang
+++ b/resources/assets/enderio/lang/en_US.lang
@@ -1121,3 +1121,7 @@ enderio.tooltip.meta=Meta:
 enderio.tooltip.uid=UID:
 enderio.tooltip.anvilupgrades=Anvil Upgrades:
 enderio.waila.config.hiddenfacades=Sneaky Facades
+
+enderio.gui.notification.poweredSpawner.bad_entity=BAD SOUL
+enderio.gui.notification.poweredSpawner.area_full=TOO MANY MOBS
+enderio.gui.notification.poweredSpawner.no_location=NO VIABLE SPACE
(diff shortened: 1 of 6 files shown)


src/main/java/crazypants/enderio/machine/spawner/TilePoweredSpawner.java
```diff
@@ -61,16 +67,53 @@ public void setSpawnMode(boolean isSpawnMode) {
...
+
   @Override
   protected void taskComplete() {
     super.taskComplete();
     if(isSpawnMode) {
       remainingSpawnTries = Config.poweredSpawnerSpawnCount + Config.poweredSpawnerMaxSpawnTries;
-      for (int i = 0; i < Config.poweredSpawnerSpawnCount && remainingSpawnTries > 0; ++i) {
-        if(!trySpawnEntity()) {
-          break;
+      TRIES: for (int i = 0; i < Config.poweredSpawnerSpawnCount && remainingSpawnTries > 0; ++i) {
```
tterrag1098: There must be a better way here than using labels and switches...

HenryLoenwind: I didn't want to redo the original loop logic. And Java uses 'break' for two different things, so...

CrazyPants:
noooo! ;P I had totally forgotten you could still even do that.

resources/assets/enderio/lang/en_US.lang
```diff
@@ -1121,3 +1121,7 @@ enderio.tooltip.meta=Meta:
 enderio.tooltip.uid=UID:
 enderio.tooltip.anvilupgrades=Anvil Upgrades:
 enderio.waila.config.hiddenfacades=Sneaky Facades
+
+enderio.gui.notification.poweredSpawner.bad_entity=BAD SOUL
+enderio.gui.notification.poweredSpawner.area_full=TOO MANY MOBS
+enderio.gui.notification.poweredSpawner.no_location=NO VIABLE SPACE
```
tterrag1098: All caps is not the convention

src/main/java/crazypants/enderio/machine/spawner/TilePoweredSpawner.java
tterrag1098: I really would prefer a cleaner approach. Using a switch nested in a for loop is ugly to begin with.

Issue #2919, opened 2015-10-21
Feature Suggestion: Magnet holds items in a buffer.


Badcholo:
In an attempt to fix lag from so many items flying around players causing the server to time out. 
Maybe if possible have the magnet hold all the items in some sort of buffer if the player can't pick them up. The player won't have access to the items until he clears inventory space or turns the magnet off.  
Other players won't have access or be able to pick up the items even though they are next to the player with the magnet and a full inventory. Items won't be flying around the player. Items would be attached to the magnet in it's buffer. Hope that makes sense. I really like the magnet. i really don't want  to ban it from the server. When the magnet is turned off the items would come out of the buffer and drop onto the floor. I understand the endless carrying capacity, but maybe it can hold Xamount of items and won't pick up any more. Then you can work on upgrades that hold more items in the buffer.
At the moment it is being used to carry endless very laggy large amounts of entities, so this maybe a good solution.

HenryLoenwind:
I think this is an X-Y problem. The real problem is that people are abusing the magnet to "carry" large amounts of items. PR #2920 will prevent that better than making the magnet an infinite pouch---which would just encourage people to do it more often.

Ultimately, a server owner has to take action against people lagging out the system with an abusive amount of item entities. I guess there are plugins that'll delete item entities if there are too many?

Badcholo:
Players will always take advantage or cause trouble for servers if they can. There is no way to watch everyone at all times 24/7 to see who is causing this. There are plugins like Clearlag that remove all dropped entities after a warning every X amount of minutes, but you can't run that commant all the time and the server gets lagged out so fast after using the magnet.
If it is possible to fix it as you suggest, I'm all for it. It sounds good and that might fix the Xp dupe.

Badcholo:
Not Enough Items mod seems to have magnet mode. When used in survival it will only pick up items that can fit in your inventory and the rest will just stay put. This seems to work very well and has a MIT license. https://github.com/Chicken-Bones/NotEnoughItems/blob/master/LICENSE.txt

Badcholo:
I did some testing and nei magnet mode works great in survival. When i remove something from my inventory to create an open space it begins to pick up items to fill that slot. I haven't been able to cause the xp dupe bug either. Maybe something like this can be used for the magnet.

Badcholo:
also it's 16 block radius for the magnet.

Fabboz:
[...]

Issue #2918, opened 2015-10-21
Magnet Issues


Badcholo:
I have experienced this issue while using the Magnet.

I have gained a lot of xp levels while mining with a mekanism atomic disassembler.
The atomic disassembler can mine incredibly fast and i am also experiencing massive server lag to the point that i time out. while i am lagged out, i can see my experience increasing very rapidly.
I saw there was an older issue but were not able to reproduce.
I am still experiencing this issue with this build.
EnderIO-1.7.10-2.3.0.417_beta
EnderCore-1.7.10-0.1.0.25_beta

I'll make a video of this as soon as i get a chance.
let me know how i can help. I don't really know how to troubleshoot this issue.
My server runs Mojangsters Modpack.
http://www.technicpack.net/modpack/mojangsters.583708
Youtube channel
https://www.youtube.com/channel/UCVRDbrSypbA696FHcaiBhkg
2 quad core cpu, 3.2, 16gig of ram.

Badcholo:
My server hard drive turns out was shit. i ran a benchmark on it and it had crapped out. Its also why i had timing out issues. I was using a 7200rpm hard drive, but am upgrading to a new SSD 240gb hard drive and i'll test out the magnet to see if my server was just running like shit. Very Sorry.

Issue #1736, opened 2015-01-23
Electromagnet Lag


devryb:
The electromagnet seems to be causing huge lag spikes and timing out most people from the server

http://puu.sh/eUaTH/accd48d50e.png

Version: 1.7.10-2.2.5.311 in the Direwolf20 1.0.3 pack

tterrag1098:
Just how many items are near the person with it on? Seems the cause here is just too many items, not really anything to do with the efficiency of the magnet.

Badcholo:
I'll prob end up disabling the Magnet cause it's causing me massive lag as well, i use the atomic disassebler from mekanism and that thing mines super fast. is there any way to craft maybe a bag that can catch all those items?

PR #2920, opened 2015-10-21
Limit number of items the magnet hoovers at a time

HenryLoenwind:
re #1736, #2918, #2919

see also #2921, which is an alternative to this one.


Badcholo:
A config to turn this up or down would be great. Some servers may want turn it up or down.

tterrag1098:
There is a config....

Badcholo:
Thank You.

Issue #2804, opened 2015-09-12
Ender IO spamming console


deadgamer1030:
I just updated to ender io  1.7.10-2.3.0.415_beta and ender core-1.7.10-0.1.0.24_beta off of jenkins and as soon I open my world the console gets spammed with this message.
[17:29:35] [Server thread/WARN] [EnderIO]: NetworkPowerManager: Exception thrown when updating power network java.lang.ArrayIndexOutOfBoundsException: 8
This happening  several times a second.
Any reason this might be happening?

Adaptivity:
https://github.com/SleepyTrousers/EnderIO/issues/1998

deadgamer1030:
maybe, it happen when updated it. maybe an outdated rf api somewhere.

hron84:
Is there a way to avoid spamming log with this message? I mean, I know there is a problem, but not everybody is able to debug and fix problems like this. But the fml logs can grow to insane sizes if you spam console with error messages in every tick.

Btw, for me this is happens if Calculator mod's Flux Point/Flux plugs works with EIO cables (in my case IIRC there is a flux point dropped to a EIO cable).

HenryLoenwind:
Not really. this is actually the program crashing hard. It's just that we abort the termination of the server and write out a message about it.

What could be done was to add the information to the message that is needed to find out what actually happened there.

hron84:
Yeah, I experienced it now, Calculator and ImmersiveEngineering does not have a protection against this and this exception is crashed my client.

HenryLoenwind:
After the change #2930 has gone live, please post one of the extended error outputs here.

hron84:
@HenryLoenwind please inform me which build contains the fix and we will try it as soon as it is possible. I have to talk with our server owner, but I think he will be open for it.

HenryLoenwind:
@hron84 That'd be the next one, 418. I'll try to remember to post in here when it's made.

hron84:
I'll keep this page open anyway.

Divineaspect:
having this as well with "crazypants.enderio.machine.generator.stirling.BlockStirlingGenerator@a55560e" being repeated endlessly.

HenryLoenwind:
Please post the log. Every tenth error message should have a full stack trace.

Divineaspect:
clarifying, fml-server-latest.log?

HenryLoenwind:
[...]

PR #2930, opened 2015-10-25
Add additional error information to logfile

HenryLoenwind:
Also reduce log spam, this error happens in tight loops.

re #2804


hron84:
@HenryLoenwind this is so great! Thanks for your quick response to this silly problem!

Issue #2862, opened 2015-09-29
Conduit/Facade visual problems with  EnderIO-1.7.10-2.3.0.415_beta


Mezzolan:
FTB Infinity just updated to  EnderIO-1.7.10-2.3.0.415_beta.jar  and I now am having issues with conduits and facades rendering properly. Conduits on their own will disappear from time to time depending on the angle of view and facades are almost completely broken - switching to and from a Yeta Wrench will usually cause them to be stuck transparent but not necessarily rendering the conduits inside. Moving from the area and returning or cycling through items repeatedly can sometimes fix the facades being stuck transparent:

![image](https://cloud.githubusercontent.com/assets/285093/10154283/fa835d34-662a-11e5-8d22-a9fdfa69de32.png) *See conduits inside facades from this angle...*

![image](https://cloud.githubusercontent.com/assets/285093/10154288/1359cfdc-662b-11e5-9b59-c331fb1bb0e2.png) *...but, turn a bit and this happens.*

![image](https://cloud.githubusercontent.com/assets/285093/10154304/29baebf8-662b-11e5-90b7-6c654f09ee03.png) *Cycling off of the Yeta Wrench in my hotbar and now it looks like this... those dark holes are actually facades over conduits like in the above images.*

Additionally, there are some anomalies with the arrows on item conduits and the like depending on the angle of viewing and whether I have the wrench out:

![image](https://cloud.githubusercontent.com/assets/285093/10154385/40aed2f6-662c-11e5-8293-d946b6393990.png) *Wrench in hand...*

![image](https://cloud.githubusercontent.com/assets/285093/10154420/a08dec70-662c-11e5-8a0b-154673022666.png) *...no wrench in hand*

FTB Infinity **does** include Fastcraft but I have the exact same issue with it active or removed from the active mods.

*(PS: The new textures look great, just can't wait for issues like this to be worked out, great job guys!)*

cloakable:
I am seeing the same problem in my modpack, no Fastcraft.

HenryLoenwind:
dupe #2812

(but thx for the screenshots, they look helpful)

Mezzolan:
No problem, hope it helps. If any more information might help feel free to ask.

SolManX:
A couple of people - including me - playing an ftb pack (infitech) are experiencing this problem. I'm on SSP but also reported on SMP.
EnderCore -> 0.1.0.24
EnderIO -> 2.3.0.415

[...]

Issue #2812, opened 2015-09-14
Bundled Conduit visual issue


BuckitHead:
After update to 415, When you walk up on bundled conduits they vanish visually. 
![2015-09-14_13 03 56](https://cloud.githubusercontent.com/assets/10180530/9860523/7feb0e04-5ae1-11e5-983f-182efc2f1927.png)

BuckitHead:
Person on the server mentioned that its not just bundled. it all works fine just visual

Sandriell:
Are you using Optifine? I got this when using it, and had to remove it.

BuckitHead:
No Optifine.

CrazyCodeUK:
I can also confirm this bug with 415 and no optifine. The conduits with disapear then reapear sometime when viewed from different angle or face away and back

tterrag1098:
For reference later this might have been broken here https://github.com/SleepyTrousers/EnderIO/commit/dfff76c0d7b240d95ff45929958f7a1d9add89ed#diff-ab2849483359c85b2ca5204960267ed3R117

tterrag1098:
I can't seem to reproduce this :/

BuckitHead:
I just updated to 416 on that sever and it's still happening... Let's me test and get back to you


Tom

BuckitHead:
ok so tried to create in single player and could not reproduce myself.  I did download the server world and tried it in single player, got the same thing so that makes this strange. Seems like the world has issues maybe. As soon as I back it down to 414 issue goes away.  I have a new just forge server I brought up a week ago, so let me update it and see what happens.

BuckitHead:
Ok on the new server it is happening again.  It's in and out now on 416,  on the other server its stayed gone on 415. I'm not sure what to tell you to help out. sorry

Reanmachine:
Playing FTB Inifinity which has 415 currently. I recorded a quick video demoing the problem:

https://www.youtube.com/watch?v=9BdO5oGfBr0

tterrag1098:
I can't test with infinity in my dev environment. I need a reproduction case with no (or a small amount of) mods, which I can actually test with.

Reanmachine:
@tterrag1098 If I get a chance this weekend ill start collapsing the infinity pack one mod at a time until i find a combination that doesn't trip the behavior and then we can go from that smaller subset or perhaps determine if it's another mod breaking ender io's behavior.

tterrag1098:
[...]

PR #2933, opened 2015-10-27
Fix render pass mixup

HenryLoenwind:
Forge stores the current render pass for blocks and tile entities in different locations. As that is not documented, it is often mixed up.

re #2862


HenryLoenwind:
also #2812

Issue #2936, opened 2015-10-29
[Suggestion] Add toggle for a blacklist on Mod Item Filter


Hypocrisy:
Right now on an insert conduit that has a Mod Item Filter on it, it will only insert items from that mod but I would like to be able to prevent items from a specific mod from passing through.

PR #2938, opened 2015-10-30
Added blacklist feature to the mod item filter

HenryLoenwind:
I'm not sure about the button placement in the GUI. Any ideas?

re #2936


mindforger:
right hand side imho, like the advanced item filter, keep it common look & feel

HenryLoenwind:
Those are not alternatives, those are the two different layouts of insert and extract.

PR #2944, opened 2015-11-01
Added Additional Hooks to make it easier to add Custom Conduits

spacebuilder2020:
Make a few modifications that would prevent crashes when trying to use a custom conduit class implementing IConduit with a custom network.

Also, made it easier to extend AbstractItemConduit for creating items for your Conduits.

### DIFF
--- a/src/main/java/crazypants/enderio/conduit/AbstractItemConduit.java
+++ b/src/main/java/crazypants/enderio/conduit/AbstractItemConduit.java
@@ -32,22 +32,30 @@ import crazypants.enderio.ModObject;
 public abstract class AbstractItemConduit extends Item implements IConduitItem {
 
   protected ModObject modObj;
+  
+  protected String unlocalisedName;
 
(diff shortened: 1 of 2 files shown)


spacebuilder2020:
I am working on a mod that will add ic2 and GT conduits, and I was running into crashes when trying to use a custom network with a custom conduit.

HenryLoenwind:
About the modObject, in Ender IO Addons I use

    public static final ModObject ModObject_blockxxx = EnumHelper.addEnum(ModObject.class, "blockxxx", new Class<?>[0], new Object[0]);

just FYI

spacebuilder2020:
Yea, the reason I used reflection in the createNetworkForType() instead of adding a new method to I Conduit is because I did not want it to break existing code.

spacebuilder2020:
Closed cause I am rewriting the createNetworkForType method based on the above suggestions.

Issue #2946, opened 2015-11-01
[bug?] Power Monitor not working


huldu:
I'm using EnderIO-1.7.10-2.3.0.417_beta and the power monitor just isn't working at all. It displays nothing at all, just 0's. Normally it would display power going in and out of the system but, nothing at all currently. It doesn't matter where I plug it in, it accepts power up until it hits the cap but like I wrote, nothing more than 0's appear.

HenryLoenwind:
It works for me. Can you give some more infos, maybe a screenshot, list of installed mods and such?

huldu:
I feel so dumb now, apparently it requires you to use conduits, I was using the power pipes from TE. My bad, sorry!

HenryLoenwind:
np, it isn't obvious. They read the data from the conduit network...

Issue #1943, opened 2015-02-16
Power Monitor not reading from Capacitor Bank anymore.


kokono2:
The Power Monitor doesn't show the the power stored and the power input and output anymore when connected directly to a Capacitor Bank. This wasn't the case a few versions before the latest (I'm not sure which version). It also doesn't read the power usage of machines directly connected to the capacitor bank (like tesseracts). It does work fine when I connect the machines and monitor to conduits.

MatthiasMann:
The power monitor always requires conduits - as it monitors devices attached to that conduit network.

Nightcaste:
I agree.  The power monitor should be able to be connected directly to a capacitor bank and still function as normal.

PR #2947, opened 2015-11-01
Display helpful text in unconnected power monitor

HenryLoenwind:
re #2946, #1943, et. al.


src/main/java/crazypants/enderio/machine/monitor/GuiPowerMonitor.java
```diff
@@ -309,6 +313,10 @@ private void renderInfoTab(int sx, int sy) {
       int sectionGap = SPACING;
 
       FontRenderer fontRenderer = getFontRenderer();
+      if (te.maxPowerInConduits == 0) {
+        fontRenderer.drawSplitString(noNetworkError, x, y, 170, ColorUtil.getRGB(Color.red));
+      } else {
+
       rgb = headingCol;
```
tterrag1098: No indents here ;P

HenryLoenwind: I didn't want to indent the whole large block because github's diff is to stupid to display it as a simple whitespace change.

Also, I forgot.

tterrag1098:
That works, I guess...

PR #2953, opened 2015-11-02
Add some more Hoes to Farming Station whitelist

### LINKED ISSUE #2952: Farming station does not accept Thermal Foundation Hoes
Hoes added by Thermal Foundation (Copper, Bronze, etc.) cannot be put into a Farming Station. I assume that this is a bug since TF Axes and Shears work without a problem.

EnderIO 1.7.10-2.3.0.417_beta
EnderCore 1.7.10-0.1.0.26_beta
CoFH Core 3.0.3
Thermal Foundation 1.2.0

wlhlm:
Closes #2952

### DIFF
--- a/src/main/java/crazypants/enderio/config/Config.java
+++ b/src/main/java/crazypants/enderio/config/Config.java
@@ -307,7 +307,11 @@ public final class Config {
       "BiomesOPlenty:hoeAmethyst", "BiomesOPlenty:hoeMud",
       "Eln:Eln.Copper Hoe",
       "Thaumcraft:ItemHoeThaumium", "Thaumcraft:ItemHoeElemental", "Thaumcraft:ItemHoeVoid",
-      "ThermalFoundation:tool.hoeInvar"
+      "ThermalFoundation:tool.hoeInvar", "ThermalFoundation:tool.hoeCopper", "ThermalFoundation:tool.hoeBronze", "ThermalFoundation:tool.hoeSilver",
+      "ThermalFoundation:tool.hoeElectrum", "ThermalFoundation:tool.hoeTin", "ThermalFoundation:tool.hoeLead", "ThermalFoundation:tool.hoeNickel",
+      "ThermalFoundation:tool.hoePlatinum",
+      "TwilightForest:item.steeleafHoe", "TwilightForest:item.ironwoodHoe",
+      "IC2:itemToolBronzeHoe"
   };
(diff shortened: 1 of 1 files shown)


wlhlm:
I've updated this PR include the Draconic Hoe, as well as some Hoes from Twilight Forest and one from IC2. I did some quick tests and they all seam to behave as expected without breaking anything (crashes, etc.).

I think this is ready.

tterrag1098:
Does this list include any unbreakable hoes? I was leaving those out on purpose.

HenryLoenwind:
The comment in the config is a bit vague. Did we leave out all unbreakables, or just those unbreakables that actually do break?

In the first case: Maybe we should have a second list with unbreakable hoes and a flag to allow them in?

wlhlm:
Hmm, not sure if it's exactly what you mean, but I just checked again, and I'm not sure about the Draconic Hoe...

The thing is, the Draconic Hoe normally runs off of RF and can be recharged. It can till a huge area all at once. When put inside a Farming Station, it works as normal, but it'll still work once its charge is depleted. I guess that makes it "unbreakable".

Should I remove it? By the time you are able to craft a Draconic Hoe (it requires killing several Withers and the Ender Dragon), does it really matter if it is unbreakable or not?

The rest of added Hoes are all "breakable".

tterrag1098:
No, it's a balance concern. people can add unbreakable hoes if they want, but I don't want them to be allowed by default.

wlhlm:
Fair enough. I removed it. The rest should be fine.

wlhlm:
BTW, how does this work for axes (and Shears for that matter)? They don't seem to use a black/whitelist, and so I'm able to put a Draconic Axe (which works similar to the Draconic Hoe) into the Farming Station and it'll still work when depleted. Why do hoes have to be balanced and Axes/Shears not?

tterrag1098:
Axes (and other tools) have methods to determine what tool type they are. Hoes aren't actually tools, they are just items that do something on right click.

HenryLoenwind:
Ah, you know how it is. One day you may decide something, but you never go back over what you did in the past and adapt it to that decision...

wlhlm:
\*nudge\* \*nudge\*

HenryLoenwind:
Here? I had to nudge tt  in IRC while hovering my mouse over the "build" button for those simple fixes yesterday. He's very busy atm, give him time. This PR is not forgotten.

PR #2959, opened 2015-11-04
Fix zero comparator signal on near-empty storage

Xyxen:
Comparators output a zero power level both when attached to empty storage, and nearly empty storage. This makes it impossible to differentiate between the two. The new semantics in this patch are equivalent to those used by Minecraft for chests, etc.

**Before:**
comparator_old

**After:**
comparator_new


Xyxen:
@tterrag1098 ?

Issue #2778, opened 2015-09-07
[Beta 413] Inverted Powered Light Bug


Xiaminou:
When I start my server all Inverted Powered Lights are off and will stay off unless I replace the power conduits on which they are (they turn on when I remove the conduit).
![2015-09-07_15 53 02](https://cloud.githubusercontent.com/assets/8670392/9718121/364e5098-5579-11e5-85e8-c480fe0c5f57.png)
![2015-09-07_15 57 17](https://cloud.githubusercontent.com/assets/8670392/9718122/36505a6e-5579-11e5-960c-928dd4d26789.png)

tterrag1098:
What version?

HenryLoenwind:
Title says Beta 413

Xiaminou:
Okay, I suppose I should have mentioned a few things:

 - I cannot reproduce the bug in a modpack with ender IO alone.
 - The bug happens in my world which also has 250+ other mods
 - The power line gets its power from a P2P Tunnel
 - It's not only on server restart but also when placing lights, I have to place the light first then the power conduit or it won't work
 - I haven't tried applying a redstone signal to it to see what happens.

Rukachan:
I can confirm using 416. You can reproduce it easily by:
1: putting down any full capacitor
2: putting down any conduit
3: putting down the inverted powered light.
The light will be fully charged, but it will not work nor it will use any RF.

I should also mention that when I break the conduit, or if I place a block next to the light it will start working again.
If the light is put directly to the capacitor, everything will work as expected.

Rukachan:
I can also confirm this in 417.

TheJulianJES:
If I put a light on a duct (from Thermal Dynamiscs). I have to update the light block first.

mindforger:
i can second @TheJulianJES it seems to miss to update itself when placed and only activates when i update the block by placings something around it

TheJulianJES:
Yeah :)

TheJulianJES:
I also have this in build 417, placing a powered light (inverted) on a duct or conduit will load the light to 100/100 RF but the light will not turn on, relogging/server restart does not fix this. So it is problay not just a render thing. So I guess it is missing something like that:
´´´
public static void onTick() {
   if(this.getEnergy() > 0) {
       world.markBlockForUpdate(x,y,z);
    }
}
´´´

TheJulianJES:
[...]

PR #2963, opened 2015-11-05
Switch on lights when they get power

HenryLoenwind:
re #2778


haXs:
I have compiled this into my version ( beta 415 ) and it does fix the issue for the wired powered lights but the issue still seems to persist for the wireless variant ( both inverted and normal ) these still require a block update to change state.

PR #2984, opened 2015-11-09
Fix: Shift-Click handling, second try

HenryLoenwind:
followup to #2411. This time really take the index into the inventory, not the slot list.

Also renamed the parameters to have a consistent naming.

### DIFF
--- a/src/main/java/crazypants/enderio/machine/gui/GuiMachineBase.java
+++ b/src/main/java/crazypants/enderio/machine/gui/GuiMachineBase.java
@@ -99,9 +99,9 @@ public abstract class GuiMachineBase<T extends AbstractMachineEntity> extends Gu
       recipeButton.visible = true;
     }
 
-    renderSlotHighlights();
-
     super.drawGuiContainerBackgroundLayer(par1, par2, par3);
+
+    renderSlotHighlights();
   }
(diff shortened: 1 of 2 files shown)


src/main/java/crazypants/enderio/machine/gui/GuiMachineBase.java
```diff
@@ -116,14 +116,15 @@ public void renderSlotHighlights() {
...
-      for (int slot = slotDef.getMinInputSlot(); slot <= slotDef.getMaxInputSlot(); slot++) {
-        renderSlotHighlight(slot, PULL_COLOR);
-      }
-    }
-    if(slotDef.getNumOutputSlots() > 0 && (mode == IoMode.PUSH || mode == IoMode.PUSH_PULL)) {
-      for (int slot = slotDef.getMinOutputSlot(); slot <= slotDef.getMaxOutputSlot(); slot++) {
-        renderSlotHighlight(slot, PUSH_COLOR);
+
+    for (Object object : inventorySlots.inventorySlots) {
+      Slot invSlot = (Slot) object;
```
tterrag1098: It's an unchecked cast either way, I prefer just casting to List<Slot>

Issue #2983, opened 2015-11-09
Server reboot after a user connects.


bakaie:
Running FTB Infinity 2.0.2 with a few extra mods installed.

We are getting a server crash and reboot after a user logs in.

http://pastebin.com/s8bJ0JkV

Looks like an issue with his farming station when he loads his chunks.  I talked with the user he has 2 farming stations, one is not powered currently, the second has power, me import bus and is farming Ender Lilies, nether wart, cotton, and a few magical crops. It currently has an amethyst hoe in it. We have several other users with Farming stations setup and working with no issue, it is just this one user.

Let me know if you need any additional information.

Thanks.

neogenix:
(I'm the user) We also tried rolling to the latest version (from a few days ago), as well as updated both EnderIO and EnderCore. It seems like when I login the chunk loads, and something odd happens. I wasn't touching the active, or inactive, station, or even in the same vicinity. The active one has been running with these plants for a few days. The only changes to it have been to add more magical crops, and remove some of the cotton, but of the same type of magical crops, no new plants.

We did have it doing essence berries, but that caused random berries, so we disabled it using the property (as it seems it's happened to others before).

mindforger:
related to https://github.com/SleepyTrousers/EnderIO/issues/2957

HenryLoenwind:
Yes, that's the same crash.

It seems that either the fake player or the world object that is part of it are not there. Really funny business; it's not as if we had a change there recently. I'll see to put some sanity checks into the next version.

HenryLoenwind:
Could you please check your logfile for the server startup: Does any mod claim to patch net.minecraft.server.management.ItemInWorldManager.func_73078_a() or net.minecraft.server.management.ItemInWorldManager.activateBlockOrUseItem()?

bakaie:
I am not seeing any in the ones that i looked at before taking off for work.  I will do a search on all the server logs later today and let you know/attach the log files if i find any.  Thanks for jumping on this and helping out.

neogenix:
Linked to duplicate #2985, closed parent.

bakaie:
Sorry for the slow response, I have looked through the log files and do not
see any of the lines you asked about.

On Mon, Nov 9, 2015 at 12:38 PM, Patrick <notifications@github.com> wrote:

> Linked to duplicate #2985
> <https://github.com/SleepyTrousers/EnderIO/pull/2985>, closed parent.
>
> —
[...]

Issue #2957, opened 2015-11-03
TInkers Mattock in farming station


cauliflower69:
Worked for a bit then crash loop 

http://pastebin.com/pB1f3KUt

HenryLoenwind:
That crash is not related to the hoe used, it happens when right-clicking a mature plant with an empty hand to harvest it.

Does anyone know what line 356 in net.minecraft.server.management.ItemInWorldManager actually is? @tterrag1098?

HenryLoenwind:
Could you please check your logfile for the server startup: Does any mod claim to patch net.minecraft.server.management.ItemInWorldManager.func_73078_a() or net.minecraft.server.management.ItemInWorldManager.activateBlockOrUseItem()?

HenryLoenwind:
The latest version contains a change that **might** help. I'd really like to get some feedback on this. Thx.

Mr-Rixa:
Tested with newest version of EnderIO and Endercore, no crash with Tinkers mattock in farming station on SMP.

PR #2985, opened 2015-11-09
Prevent FarmerJoe from being removed from the world

HenryLoenwind:
This might be the cause of #2957 and #2983. About the only plausible explaination for the crash, unless someone is patching activateBlockOrUseItem() badly.

### DIFF
--- a/src/main/java/crazypants/enderio/machine/farm/FakeFarmPlayer.java
+++ b/src/main/java/crazypants/enderio/machine/farm/FakeFarmPlayer.java
@@ -17,6 +17,7 @@ import net.minecraft.world.WorldServer;
 import com.mojang.authlib.GameProfile;
 
 import cpw.mods.fml.common.FMLCommonHandler;
+import crazypants.enderio.Log;
 
 public class FakeFarmPlayer extends EntityPlayerMP {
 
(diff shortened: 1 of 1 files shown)


wlhlm:
Didn't know the farmer had a name...  ;-)

mindforger:
probably Fastcraft related? Both Crash reports state FastCraft as installed!

HenryLoenwind:
Could be. But until we get an idea what actually is null, we're guessing. The method is crashing gets three parameters that can be null (and a couple of ints which cannot be). The fake player, the fake player's world object and an itemstack. The itemstack can and is null (=empty hand). The fake player cannot be null in our code path. That leaves the world object. It must be non-cull when the fake player is created (or it would crash then) and it has only one method where it is assigned in the fake player's class tree. The one I modified above. Ok, someone could just assign the field directly, but I cannot catch that without some major patching.

src/main/java/crazypants/enderio/machine/farm/FakeFarmPlayer.java
```diff
@@ -88,4 +89,12 @@ public void func_147100_a(C15PacketClientSettings pkt) {
   public boolean canPlayerEdit(int par1, int par2, int par3, int par4, ItemStack par5ItemStack) {
     return true;
   }
+
+  @Override
+  public void setWorld(World p_70029_1_) {
+    Log.warn("Ender IO Farming station fake player is being transfered to world '" + p_70029_1_
+        + "'. Trying to reject transfer. Call stack follows:");
+    (new Exception()).printStackTrace();
```
tterrag1098: `Thread.dumpStack()`

HenryLoenwind: F3

```
	public static void dumpStack() {
		new Exception("Stack trace").printStackTrace();
	}
```

ooookeeey... ;)

tterrag1098: Heh, I didn't actually read the source, but the it's more about the principle, no? :P

Issue #2196, opened 2015-03-31
Crash with Travel Anchor


Niduin:
Minecraft crashed when i right click an Travel Anchor.

Time: 31.03.15 21:41
Description: Unexpected error

java.lang.NullPointerException: Unexpected error
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:213)
	at net.minecraftforge.common.UsernameCache.getLastKnownUsername(UsernameCache.java:101)
	at crazypants.enderio.teleport.anchor.BlockTravelAnchor.sendPrivateChatMessage(BlockTravelAnchor.java:150)
	at crazypants.enderio.teleport.anchor.BlockTravelAnchor.openGui(BlockTravelAnchor.java:142)
	at crazypants.enderio.BlockEio.func_149727_a(BlockEio.java:100)
	at net.minecraft.client.multiplayer.PlayerControllerMP.func_78760_a(PlayerControllerMP.java:326)
	at net.minecraft.client.Minecraft.func_147121_ag(Minecraft.java:1435)
	at net.minecraft.client.Minecraft.func_71407_l(Minecraft.java:1941)
	at net.minecraft.client.Minecraft.func_71411_J(Minecraft.java:961)
	at net.minecraft.client.Minecraft.func_99999_d(Minecraft.java:887)
	at net.minecraft.client.main.Main.main(SourceFile:148)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at net.minecraft.launchwrapper.Launch.launch(Launch.java:135)
	at net.minecraft.launchwrapper.Launch.main(Launch.java:28)

A detailed walkthrough of the error, its code path and all known details is as follows:
---------------------------------------------------------------------------------------

-- Head --
Stacktrace:
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:213)
	at net.minecraftforge.common.UsernameCache.getLastKnownUsername(UsernameCache.java:101)
	at crazypants.enderio.teleport.anchor.BlockTravelAnchor.sendPrivateChatMessage(BlockTravelAnchor.java:150)
	at crazypants.enderio.teleport.anchor.BlockTravelAnchor.openGui(BlockTravelAnchor.java:142)
[...]

PR #3013, opened 2015-11-19
Prevent unowned travel anchor crashing

HenryLoenwind:
Still doesn't fix the problem that they can go unowned...

re #2196


HenryLoenwind:
This has been superceded by the "wrap UUID" PR and is no longer needed.

Issue #2102, opened 2015-03-18
Only null may change this block's settings


PsionicArchon:
Using the latest jenkins release I am no longer able to change the travel anchor's settings.

Example: http://i.imgur.com/dIsec8m.png

Tsunayoshi:
I have the same problem. But I am not getting any messages. Click an old Anchor (placed in the EnderIO version before) the game crashes: http://pastebin.com/0LKs2XAX

shady2k:
Same problem here. Than right click on anchor "Only null may change this block settings" appeared.

Soundtoxin:
I am also experiencing this issue on the latest version of EnderIO, and I experienced it on one several versions back. I take great pride in the neat labeling of my Travel Anchors, so I hope this bug is considered high priority.

PR #3016, opened 2015-11-19
Wrap UUID

HenryLoenwind:
UUID handling is now done by an object that knows how to deal with it. It also knows what to do when some of the data belongs to an offline player. It also records user names, so they are available later even if the game doesn't have them.

re #2102, #2196


HenryLoenwind:
Ok, tt. I added the username refresh on load.

Issue #3024, opened 2015-11-21
Obelisks not rendering correctly when in hotbar


Native91:
All obelisks don't seem to render correctly when in the hotbar. (http://i.imgur.com/Z43sUYg.png)

They do seem to render fine when the inventory is opened (http://i.imgur.com/nHco3f2.png)

HenryLoenwind:
Which version of Ender IO?

wlhlm:
I can observe similar issues:
![2015-11-21_22 40 13](https://cloud.githubusercontent.com/assets/1928113/11320665/52cfa056-90a1-11e5-919a-18d2e103261a.png)

__How to reproduce__
- switch to creative game mode
- drop Obelisk in hotbar
- switch between tabs in creative inventory or search for something

EnderIO 2.3.0.420
EnderCore 0.2.0.31
Forge 1558

HenryLoenwind:
Oh I hate item rendering sometimes. It is like driving a Formula 1 car with a blindfold on a beach. You never know what hits you, could be a parasol, could be a seal, could be a Navy Seal.

What happens here is that some other item that is rendered before disables transparency completely. So the box that is there for the bottom texture has all its transparency rendered as white.

PR #3027, opened 2015-11-22
Cleanup obelisk rendering even more

HenryLoenwind:
and make sure we have the right OpenGL bits set when rendering as item. This should also fix some older rendering issues with the floating item transparency.

re #3024


tterrag1098:
Thanks, I could never quite figure out what was causing the item rendering issue.

Issue #3026, opened 2015-11-22
alloy smelter only smelting one item at a time


theboo89:
so im having an issue with the alloy smelter... say im wanting to make quite clear glass, if i put 3 stacks of sand in, it smelts 3 at a time like it always has. But if you just put 1 stack in a single slot, it only smelts one at a time, i dont know if this was intentional or not, but it creates a problem because when you put 3 stacks of sand in to the smelter, it only pulls sand from the left hand slot, and when it runs out it starts pulling from the middle slot, but since only 2 slots have sand, it is now only smelting 2 at a time, and when thats gone it pulls from the right slot and only smelts one at a time... seems kinda weird

im using the 417 beta on my server, but i just loaded 420 in single player with no other mods and i am having the same issue.

i can understand if you want it to only smelt 3 at a time when all 3 slots are filled, but if were gonna have that, it should pull equally from all 3 slots ...

thanks!
Tristan 

edit: Forgot to mention, i looked up and down throughout the configs and couldnt find anything for this, so hopefully im not just blind haha

edit2: i think it all evens out in the end because it scaled the time based on how many blocks its smelting, but it still seems weird that it works this way

HenryLoenwind:
The Alloy Smelter only smelts 3 at a time for furnace recipes (vanilla glass), not alloy smelter recipes (quite clear glass). It also pulls input items from the lowest slot is can find them. Always been that way.

theboo89:
i c... well it does smelt 3 quite clear glass at a time if all 3 slots have items.. so i dont know if thats intentional or not... anyways, thx for the response

HenryLoenwind:
Oh, someone added 2 additional recipes to the config some time ago. One that takes 2*1 sands and one that takes 3*1. There are none for 3*1, 1*1+2*1 and 2*1+1*1.

Intentional? Not really, those additional recipes are there because the normal recipe actually reads "1 slot with 1 or more sand, 2 empty slots"---and would fail if there were 3 slots of sand. It was a hotfix back then and never got improved.

Issue #1070, opened 2014-10-06
Alloy smelter produces normal Glass with 2-3 Stacks of Sand


Xaw4:
when there are more than one input slot is filled in the Alloy smelter, it will switch to "normal Smelting" and make vanilla glass, if in "All Smelting" mode.
i am not sure if this is intended or not. but it buggs me out a bit, that i get vanilla glass out of my sag mill - smelter combo, when i throw more then one stack in.. and i dont want to switch around the mode button every time.. because it works perfect in all the other instances.

PR #3030, opened 2015-11-22
Improve handling of single-input recipes

HenryLoenwind:
* Single input alloy smelter recipes now will get multi-slot variants
added automatically
* There's a config flag to disable this
* Removed existing partila multi-slot recipes
* Fixed code that depends on the order of recipes in a HashMap
* Fixed code that depends on the order of machineRecipes in a HashMap

re #3026, #1070

Note: We still have a problem with recipes that need only a single item but more than 1 of it. So the 4xnetherQuartz one will fail on an input of "64/1/0". And if there is a vanilla recipe available for smelting netherQuartz, it will take over.

To fix this, we'd need to replace large parts of the recipe system.

### DIFF
--- a/resources/assets/enderio/config/AlloySmelterRecipes_Core.xml
+++ b/resources/assets/enderio/config/AlloySmelterRecipes_Core.xml
@@ -145,50 +145,12 @@
         <itemStack modID="EnderIO" itemName="blockFusedQuartz" itemMeta="0" exp="0.5" />
       </output>
     </recipe>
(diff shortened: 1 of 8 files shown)


tterrag1098:
That many extra recipes is going to destroy lookup performance. Not sure this is worth it?

HenryLoenwind:
It's not that many extras. The default config has 3 recipes that match this, and they already had some multi-slot variants in the xml that I removed. So it's about 10 recipes more than before. Also compared to the lookup "is it a smelting recipe" that is nothing.

Issue #3041, opened 2015-11-28
Soul Binder & Slice'n'Splice recipes backgrounds...


therealericbj:
Using EnderIO-1.7.10-2.3.0.421_beta.jar, I discovered this...

![slice-n-splice](https://cloud.githubusercontent.com/assets/8407817/11450986/0a4f8dae-957a-11e5-8a11-1502878b585c.png)

![soul-binder](https://cloud.githubusercontent.com/assets/8407817/11450987/111a8206-957a-11e5-92a2-e4705195addd.png)

I don't know what else I can tell you about it.  There are no errors in the console, no crashing...

If there are any other details I can get for you, just say so...

HenryLoenwind:
Ooops, my bad. I broke it.

XFRGod:
Herpa-Derp?

therealericbj:
Mister Henry, could this be part of the same problem?

![rr-01](https://cloud.githubusercontent.com/assets/8407817/11459107/9beaaac6-9694-11e5-8f5f-83853978c176.png)

![rr-02](https://cloud.githubusercontent.com/assets/8407817/11459111/ab77ebac-9694-11e5-8373-cdf05ed083c0.png)

![rr-03](https://cloud.githubusercontent.com/assets/8407817/11459113/b782fff4-9694-11e5-9fb7-a0d08d9f21ad.png)

In the third picture, you can see most things are showing up quite nicely, but the Machine Chassis and the Conduit Facade seem a little off.

The storage mod that is being used is Router Reborn, RouterReborn-1.7.10-1.2.0.35-universal.jar.

HenryLoenwind:
No, that is something else. Somehow the wrong texture sheet is used when rendering. I'll have a look in our code, but I suspect that Router Reborn isn't doing the same texture setup as vanilla.

PR #3046, opened 2015-11-30
Ensure correct texture is used for conduit items

HenryLoenwind:
re #3041


tterrag1098:
The item sheet will already have been bound at that point. I'm not sure what this accomplishes.

Issue #1530, opened 2014-12-30
Insane Amounts of XP EnderIO -1.7.10-2.2.6.313_beta


SJ1133:
When ever i manually put in any dusts to be smelted in the alloy smelter in furnace mode and then pick it up once it has smelted up - 3 ingots picked up == to 17 LEVELS.
XP gain increases the more ingots i pick up at in one go from the smelter
EDIT: Seems to be happening with any ore from thermal foundation even if it is directly smelted,

mmelvin0:
I ran into this and think I may have randomly stumbled upon... something.

XP from the Alloy Smelter has always been a bit wonky for me, but it seems like it either gave the correct amount, or absolutely nothing. However, smelting a stack of birch logs into charcoal gives me 32 levels consistently. I tried a number of things to work around it, but without luck.

This was only happening to me with a dedicated server. Single player worked as expected.

Then I noticed I was getting a debug message about the following in my `OreDictionaryPreferences_User.xml` file:

```
<preference oreDictionary="ingotAluminuim">
  <itemStack modID="TConstruct" itemName="materials" itemMeta="11" />
</preference>
```

It seems by correcting ingotAlumin**ui**m to ingotAlumin**iu**m I get the same amount of XP from birch logs as smelting in a normal furnace.

I have absolutely no idea if this is the solution or a red herring (esp. since there are TONS of similar debug message about non-existent ore dictionary names from `SAGMillRecipes_Core.xml`) but I figure I'd mention it since it (seems to have) worked for me and it all shares the same base parsing code (although I realize `OreDictionaryPreferenceParser` is responsible for this and shouldn't effect vanilla smelting recipes.)

This is with EnderCore 0.2.0.31_beta and EnderIO 2.3.0.421_beta.

HenryLoenwind:
@mmelvin0 Can you reproduce this again? Would you be willing to do so with a debugging version of Ender IO?

mmelvin0:
I'm positive I can reproduce it with the current (soon to be previous) version of my mod pack.

If you have a debugging version for me to try (I assume this means you have some ideas...) I'd be more than happy to try.

I already have an EnderIO dev environment set up from a while back. Was planning on stepping through the code later this week to see if anything jumps out at me, but your suggestion sounds like far less effort :)

HenryLoenwind:
[...]

PR #3062, opened 2015-12-14
Limit smelting XP the same way vanilla does (thx to mmelvin0)

HenryLoenwind:
re #1530

Root cause: Some mods (e.g. Biomes of Plenty) register smelting recipes with invalid XP multipliers. Vanilla ignores them.


immibis:
Vanilla limits you to 1 XP per smelted item?

Sounds more like a bug in vanilla, really.

HenryLoenwind:
From the code and the values I'd say Vanilla sees it as percentage; the change to get an XP for smelting an item.

Issue #3069, opened 2015-12-17
FakeNetHandlerPlayServer error


MySt1k:
http://pastebin.com/x5JMvR7w

HenryLoenwind:
```
EnderIO: Found the following problem(s) with your installation:
                 An unsupported base software is installed: 'kcauldron, cauldron, craftbukkit, mcpc'. This is NOT supported.
                 This may have caused the error. Try reproducing the crash WITHOUT this/these mod(s) before reporting it.
```

If you get me the source of the NetHandlerPlayServer class that is used in your environment, I'll have a look if I can work around it. Otherwise: Ender IO works with the Minecraft classes as they are shipped by Mojang and patched by Forge.

Calenria:
https://github.com/spannerman79/KCauldron/blob/master/patches/net/minecraft/network/NetHandlerPlayServer.java.patch

PR #3071, opened 2015-12-17
Prevent outright crash when used on a kcauldron server

HenryLoenwind:
re #3069


tterrag1098:
Might as well, even though KCauldron has fixed this on their end.

PR #3081, opened 2015-12-20
Check if chunk is loaded in farming station

HenryLoenwind:


src/main/java/crazypants/enderio/machine/farm/TileFarmStation.java
```diff
@@ -408,13 +410,14 @@ protected void doTick() {
       clearNotification();
     }
 
-    BlockCoord bc = getNextCoord();
-    if(bc != null && bc.equals(getLocation())) { //don't try and harvest ourselves
+    BlockCoord bc = null;
+    int infiniteLoop = 20;
+    while (bc == null || bc.equals(getLocation()) || !worldObj.getChunkProvider().chunkExists(bc.x >> 4, bc.z >> 4)) {
```
tterrag1098: Just use blockExists...

HenryLoenwind: Why should I waste CPU cycles on checking y for 0-255 and another method call?

doChunksNearChunkExist() however, might be  a good idea to prevent block updates from spilling over.

Issue #3170, opened 2016-01-31
Zombie Generator permanently stops when full


CyricV:
I'm running into what I believe is an error with the Zombie Generator and I'd like to know if there is a bug I should report or if I'm messing things up myself.

The Generator works fine until it fills up both it's stored energy and Nutrient Distillation. At this point it stops generating energy and will not start again. The 100K RF buffer can be emptied out just fine, but the Generator will sit with 2000/2000 mB of Nutrient Distillation and 0/100000 RF doing nothing. As long as the generator never maxes out both Nutrient Distillation and RF it's fine, but once that happens it becomes stuck and must be broken and replaced.

Ender IO 2.3.0.422
EnderCore 0.2.0.31

HenryLoenwind:
Can you please check that you didn't set a redstone mode on the generator?

CyricV:
Confirmed, I've tested with all different redstone modes on independent generators and they all have the same issue. I'm going to do a quick test with only Ender IO installed to see if the issue is related to other mods.

HenryLoenwind:
Thx for doing so, in my dev env (with minimal mods...) it doesn't happen.

CyricV:
OK, I think I've figured it out, part of the trouble I had testing was a non default config following me around between Minecraft instances, but discovering that actually helped me sort the issue.

It seems that when both the RF and Distillation are maxed on a Zombie Generator it goes into a kind of standby, and to wake it up again you need to lower the stored RF by a certain amount. For the default 80 RF/T happens when you drop it to about 90K RF in the buffer. However, if you set the generation rate higher that also seems to increase the amount you need to pull to wake it back up. At 400 RF/T it must be dropped to about 20K before it wakes up and with 600 RF/T it cannot be woken back up even if completely drained. I'm not sure if this is intended behavior or just a weird artifact that no one has found or cared to mention since I seem to be the only person who likes the Zombie Generator anyways. I'm going to see what I can set the max RF/T to before it stops being able to function.

CyricV:
Seems like 499 RF/T is the cutoff point, so at 500 RF/T the generator will not wake up after going into standby.

HenryLoenwind:
It says:

```
    //once full, don't start again until we have drained 10 seconds worth of power to prevent
    //flickering on and off constantly when powering a machine that draws less than this produces
```

Yeah, that'd be your 500*20*10 >= 10000

CyricV:
[...]

PR #3171, opened 2016-01-31
Prevent Zombie Gen from being configured to death

HenryLoenwind:
* Too high rf/t will no longer stall it (value will be decreased instead)
* Resume time can be configured (so rf/t can be configured higher than 500)
* Config values can be changed at runtime
* re #3170

(pun intended)

### DIFF
--- a/src/main/java/crazypants/enderio/machine/generator/zombie/TileZombieGenerator.java
+++ b/src/main/java/crazypants/enderio/machine/generator/zombie/TileZombieGenerator.java
@@ -31,9 +31,6 @@ public class TileZombieGenerator extends AbstractGeneratorEntity implements IFlu
 
   final SmartTank fuelTank = new SmartTank(EnderIO.fluidNutrientDistillation, FluidContainerRegistry.BUCKET_VOLUME * 2);
 
-  int outputPerTick = Config.zombieGeneratorRfPerTick;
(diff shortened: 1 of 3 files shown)


src/main/java/crazypants/enderio/machine/generator/zombie/TileZombieGenerator.java
```diff
@@ -146,7 +143,7 @@ private boolean generateEnergy() {
 
     //once full, don't start again until we have drained 10 seconds worth of power to prevent
     //flickering on and off constantly when powering a machine that draws less than this produces
-    if(inPause && getEnergyStored() >= (getMaxEnergyStored() - (outputPerTick * 200))) {
+    if (inPause && getEnergyStored() > (getMaxEnergyStored() - (getPowerUsePerTick() * Config.zombieGeneratorResumeTime))) {
```
UberWaffe: Would it not make sense here to do something like check for less than the '10 seconds anti-chatter' OR 'completely empty'?
`boolean energyLow = (getEnergyStored() > (getMaxEnergyStored() - (getPowerUsePerTick() * Config.zombieGeneratorResumeTime))) || (getEnergyStored() <= 0)
if (inPause && energyLow ) {`

(Note: Possible syntax mistakes, I don't use java on a regular basis).

HenryLoenwind:
#3601 solves this problem in a much simpler way

PR #3210, opened 2016-02-26
Update zh_CN.lang

3TUSK:
I just found two inappropriate translations.
* Translation for Dense ME Conduit should match [the style of that one in AE](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/master/src/main/resources/assets/appliedenergistics2/lang/zh_CN.lang#L442).
* Shovel ("铲") -> Pick/Pickaxe ("镐").


3TUSK:
Still no message? Or I pushed into wrong branch?

Issue #3096, opened 2015-12-26
Advanced photovoltaic cell max capacity grows when partially unloaded


Eppiox:
If advanced photovoltaic cell are unloaded and reload with chunk-loaded cells, the original cluster value goes up by X amount, where X is the amount of unloaded cells.
I have 74 cells(740,000 rf max capacity), 17 are unloaded- if I reload the chunk they are in the new max capacity is 910,000 rf.  (740,000 + 170,000)

Example with 74-> http://imgur.com/nhBMWtZ  (further chunk loaded, chunk containing player and 17 cells not permanently loaded)

Exploit example-> http://imgur.com/DPbdTjB

PR #3215, opened 2016-03-01
Reworked Solar Panel network

HenryLoenwind:
- Fix: Detect chunk-unloaded panels
- Fix: Save/load energy to/from save file
- Fix: Stop chunkloading neighboring chunks
- General cleanup

re #3096

### DIFF
--- a/src/main/java/crazypants/enderio/machine/solar/SolarPanelNetwork.java
+++ b/src/main/java/crazypants/enderio/machine/solar/SolarPanelNetwork.java
@@ -1,18 +1,22 @@
 package crazypants.enderio.machine.solar;
 
 import java.util.EnumSet;
+import java.util.Iterator;
 import java.util.List;
 
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.tileentity.TileEntity;
+import net.minecraft.world.World;
 import net.minecraftforge.common.util.ForgeDirection;
 import cofh.api.energy.EnergyStorage;
 
+import com.enderio.core.common.util.BlockCoord;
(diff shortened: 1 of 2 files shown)


src/main/java/crazypants/enderio/machine/solar/SolarPanelNetwork.java
```diff
@@ -61,21 +89,53 @@ boolean addToNetwork(TileEntitySolarPanel panel) {
     return true;
   }
 
+  void cleanupMemberlist() {
+    if (!panels.isEmpty()) {
```
firios: Consider reverting the direction of these if clauses.
like:
if(panels.isEmpty()) {
  return;
}

Having less layers of indentation will make this easier to read

src/main/java/crazypants/enderio/machine/solar/SolarPanelNetwork.java
```diff
@@ -61,21 +89,53 @@ boolean addToNetwork(TileEntitySolarPanel panel) {
...
+        boolean isGood = false;
+        if (world.blockExists(panel.x, panel.y, panel.z)) {
+          TileEntity tileEntity = panel.getTileEntity(world);
+          if (tileEntity instanceof TileEntitySolarPanel) {
+            if (((TileEntitySolarPanel) tileEntity).network == this) {
+              isGood = true;
+            }
+          }
+        }
+        if (!isGood) {
```
firios: Could the variable be isBad or similiar instead? the ! is easy to miss when reading

PR #3250, opened 2016-04-02
Fixed #3249

### LINKED ISSUE #3249: RF connection not working with IEnergyReceiver and IEnergyProvider on the same tile entity
#### Issue Description:

EnderIO power interface is making invalid assumption on CoFH Energy blocks
#### What happens:

When connecting with a tile entity implementing both IEnergyReceiver and IEnergyProvider, EnderIO only check for the latest and doesn't realize that the same tile entity can also receive energy.
As seen here: https://github.com/SleepyTrousers/EnderIO/blob/master/src/main/java/crazypants/enderio/power/PowerHandlerUtil.java#L21

EnderIO should support tile entities with both direction of RF energy.
#### Steps to reproduce:
1. place a WarpDrive energy bank
2. fill the bank with energy by placing a EnderIO creative capacitor next to it
3. empty the bank with energy by placing an empty EnderIO capacitor next to it

---
#### Affected Versions (Do _not_ use "latest"):
- EnderIO: EnderIO-1.7.10-2.3.0.424_beta-dev
- EnderCore: EnderCore-1.7.10-0.2.0.31_beta-dev
- Minecraft: 1.7.10
- Forge: 10.13.4.1490

N/A

LemADEC:


MatthiasMann:
What's the point of this change?

tterrag1098:
@MatthiasMann Some blocks may implement IER and IEP separately, instead of using IEH. In 1.8 this is the recommended way, so we should probably adapt.

PR #3287, opened 2016-05-02
Show Block Name in GuiExternalConnectionSelector. Fixes #1984

### LINKED ISSUE #1984: Suggestion: Yeta Wrench: show TileEntity Names when selecting direction.
It would be awesome, if the GUI to select the direction of the Conduit you want to edit (North, South, West, East, Up, Down) would also show the name of the Block/TileEntity,  that the conduit is connected to in that direction.

For most of the cases where you don't connect the same blocktypes with eachother this would make it easier to figure out which direction you need to edit without needing to look at the map/F3 mode.

MaPePeR:
Its been a year, since #1984, so i thought i might give it a shot.

image

For 1.7.10. I hope you still accept PRs for it and that it can be easily ported to 1.9.8 and above.

### DIFF
--- a/src/main/java/crazypants/enderio/conduit/gui/GuiExternalConnectionSelector.java
+++ b/src/main/java/crazypants/enderio/conduit/gui/GuiExternalConnectionSelector.java
@@ -5,11 +5,13 @@ import java.awt.Point;
 import java.util.HashSet;
 import java.util.Set;
 
+import net.minecraft.block.Block;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.entity.EntityClientPlayerMP;
 import net.minecraft.client.gui.GuiButton;
 import net.minecraft.client.gui.GuiScreen;
 import net.minecraft.client.renderer.Tessellator;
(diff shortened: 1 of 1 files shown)


tterrag1098:
@HenryLoenwind Will your changes to the GUI in 1.8 easily backport to this version? If not this PR seems like a good midway point.

MaPePeR:
Are there changes in the 1.8 version that make my changes useless? 

If not I will probably write another pull for 1.8 this evening. 

I'm really sick of seeing DW20 using F3 to figure out the direction :wink:

HenryLoenwind:
"I'm really sick of seeing DW20 using F3 to figure out the direction"---LOL, that was what made me do the 1.8.9 changes. I literally paused the video to do it.

I think those two changes will combine nicely. But we may want to reduce the text size of the block names. (That should be possible somehow, the tooltip of the "Enderios" does it...)

MaPePeR:
See #3288 for the 1.8.9 Pull Request. I could not find out how to easily reduce the font size, but i think it might be fine without that.

PR #3288, opened 2016-05-03
Show Block Name in GuiExternalConnectionSelector in 1.8.9. 

MaPePeR:
Implemens #1984 
See #3287 for 1.7.10 PR.

### DIFF
--- a/src/main/java/crazypants/enderio/conduit/gui/GuiExternalConnectionSelector.java
+++ b/src/main/java/crazypants/enderio/conduit/gui/GuiExternalConnectionSelector.java
@@ -11,7 +11,7 @@ import net.minecraft.client.Minecraft;
 import net.minecraft.client.entity.EntityPlayerSP;
 import net.minecraft.client.gui.GuiButton;
 import net.minecraft.client.gui.GuiScreen;
-import net.minecraft.init.Blocks;
+import net.minecraft.util.BlockPos;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.MathHelper;
 
@@ -24,6 +24,7 @@ import crazypants.enderio.conduit.IConduit;
 import crazypants.enderio.conduit.IConduitBundle;
 import crazypants.enderio.conduit.redstone.IInsulatedRedstoneConduit;
 import crazypants.enderio.network.PacketHandler;
+import net.minecraft.world.World;
 
 public class GuiExternalConnectionSelector extends GuiScreen {
 
@@ -85,8 +86,13 @@ public class GuiExternalConnectionSelector extends GuiScreen {
   }
 
   protected String getBlockNameForDirection(EnumFacing direction) {
-    Block b = cb.getBundleWorldObj().getBlockState(cb.getLocation().getLocation(direction).getBlockPos()).getBlock();
-    if (b != null && b != Blocks.air) {
+    World world = cb.getBundleWorldObj();
+    BlockPos blockPos = cb.getLocation().getLocation(direction).getBlockPos();
+    if (world.isAirBlock(blockPos)) {
+      return null;
+    }
+    Block b = world.getBlockState(blockPos).getBlock();
+    if (b != null && b != EnderIO.blockConduitBundle) {
       return b.getLocalizedName();
     }
     return null;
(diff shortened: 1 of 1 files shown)


src/main/java/crazypants/enderio/conduit/gui/GuiExternalConnectionSelector.java
```diff
@@ -80,11 +84,21 @@ private void go(EnumFacing dir) {
     player.openGui(EnderIO.instance, GuiHandler.GUI_ID_EXTERNAL_CONNECTION_BASE + dir.ordinal(), player.worldObj, loc.x, loc.y, loc.z);
   }
 
+  protected String getBlockNameForDirection(EnumFacing direction) {
+    Block b = cb.getBundleWorldObj().getBlockState(cb.getLocation().getLocation(direction).getBlockPos()).getBlock();
+    if (b != null && b != Blocks.air) {
```
HenryLoenwind: Please use world.isAirBlock(), otherwise things like the LightNode will show up here, too. I'd also like conduits filtered out, too.

src/main/java/crazypants/enderio/conduit/gui/GuiExternalConnectionSelector.java
```diff
@@ -27,6 +29,8 @@
 
   Set<EnumFacing> cons;
   IConduitBundle cb;
+  String adjacentBlockNames[] = new String[6];
```
HenryLoenwind: Plase use Enumfacing.values().length, or even better EnumMap<EnumFacing,String> instead of the array.

src/main/java/crazypants/enderio/conduit/gui/GuiExternalConnectionSelector.java
```diff
@@ -111,6 +125,14 @@ public void drawScreen(int par1, int par2, float par3) {
     int y = height / 2 - butHeight * 3 - 5;
         
     drawString(Minecraft.getMinecraft().fontRendererObj, txt, x, y, ColorUtil.getARGB(Color.white));
+    for (int i = 0; i < 6; i++) {
+      String blockName = adjacentBlockNames[i];
+      if (blockName == null) {
+        continue;
+      }
+      int textWidth = Minecraft.getMinecraft().fontRendererObj.getStringWidth(blockName);
+      drawString(Minecraft.getMinecraft().fontRendererObj, blockName, textPositions[i].x + 60 / 2 - textWidth / 2, textPositions[i].y, ColorUtil.getARGB(Color.gray));
```
HenryLoenwind: Can you move this up, so it is rendered before ("behind") the buttons? There is a chance of overlap here.

HenryLoenwind:
Oh, and something I noticed on your screenshot: Your text is touching the buttons, it needs to move down 1 pixel.

MaPePeR:
image

Text is now 1 pixel lower and rendered behind buttons. Didn't look like i had to do some GL-stuff.

HenryLoenwind:
Is it just that image or is the background a solid black now?

MaPePeR:
Its just the background. if you look VERY closely you can see a coal ore on the right. I think i just took the screenshot at night.

HenryLoenwind:
ok, that's fine. It may also be that the monitors at work are not really configured for anything. Actually, I know they aren't---I set them up. ;)

HenryLoenwind:
2 things: First, Block.getUnlocalizedName() is notoriously unreliable. It doesn't even work for all vanilla blocks, let alone modded blocks. Second, I expanded it a bit, I hope you don't mind:

MaPePeR:
It's great. Thanks a lot  for improving it! Especially that late at night. 

It's now even better, than I could have imagined.

HenryLoenwind:
That's the wonder of collaboration; the result is better than what any participant could have accomplished on their own.

But I'm still not sure if we are making clear what each button represents. So far we only have:

* Its position
* The name of the direction
* The name of the block
* The picture of the block

Maybe we should add a couple more hints....lol, just kidding ;)

tterrag1098:
I think it looks best with normal sized text in the buttons, and unicode text below.

PR #3433, opened 2016-06-21
Me dense conduit and me conduit

Nico836:
I think that it is a real pity that we cannot color cables as with those of ae. I would want to be able to color them to have no more conflicts and be also able to find me between everything cables there.


MatthiasMann:
This is not an issue report ...

HenryLoenwind:
...and if it was, it would be a duplicate.

PR #3446, opened 2016-06-25
Improve Build Times

RlonRyan:
Updated the bundled Gradle wrapper to Gradle 2.14. Local testing shows
build improvements of up to 20 seconds, an approximate 25% build time
reduction. On average, the build time differences are nominal, but
several other improvements have been made to the Gradle system since
version 2.0, including buffs to IDE compatibility. Additionally, added
the option to have Gradle login to the developer's Minecraft account by
setting the variables `minecraft_username` and `minecraft_password` in
the `${userhome}/.gradle/gradle.properties` file. This is to encourage
users to develop using valid Minecraft accounts. It is important to take
care as not to set the fields in the project's local gradle.properties
file, as to avoid commiting your Minecraft account information to the
public. Finally, added .gitignore entries for NetBeans related files, to
prevent NetBeans users from committing IDE files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ language: java
 # Tell Travis this is a Java 7+8 project.
 matrix:
   include:
-    - jdk: oraclejdk7
+    - jdk: openjdk7
     - jdk: oraclejdk8
 
 # Tell Travis we want container-based builds.


tterrag1098:
We test on openjdk and oraclejdk for a reason. Why change that?

RlonRyan:
OracleJDK is the standard JDK for java. The more important difference is the difference between Java 7 and Java 8 compiles. If you would prefer, I would be more than happy change the Java 7 JDK back to OpenJDK 7.

tterrag1098:
The server the mod is compiled on uses openjdk8. However travis does not have this available (yet) so we compromised with openjdk7. As far as the differences between different oracle jdk versions, I think it is negligible and not worth testing every PR against.

RlonRyan:
I see... Arguably it is not worth testing against more than one JDK version, but for the sake of security I will switch back the jdk7.

tterrag1098:
Ok, I will let @HenryLoenwind review though, since he is the one who initially set up travis.

RlonRyan:
Sounds good! I hope these modifications will work well for you, as they have for me in the past.

.travis.yml
```diff
@@ -1,11 +1,53 @@
...
+before_cache:
+  - rm -f $HOME/.gradle/caches/modules-2/modules-2.lock
+  - rm -f $HOME/.gradle/caches/minecraft/ForgeVersion.json
+  - rm -f $HOME/.gradle/caches/minecraft/ForgeVersion.json.etag
+  - rm -rf $HOME/.gradle/caches/minecraft/deobfedDeps
+
+# Cache libs, as to improve speed.
 cache:
   directories:
-    - '$HOME/.m2'
```
HenryLoenwind: why?

.travis.yml
```diff
@@ -1,11 +1,53 @@
+# Tell Travis which branches to build.
+branches:
```
HenryLoenwind: that doesn't fit out git setup at all

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Aug 19 17:12:51 CDT 2014
```
HenryLoenwind: this (and below) will be overwritten the next time we update Forge

build.gradle
```diff
@@ -69,11 +69,11 @@ dependencies {
     compile ("com.pahimar.ee3:EquivalentExchange3:${ee3_version}:dev") {
         transitive = false
     }
-
```
HenryLoenwind: all those whitespace changes will make it a pain to merge this into the 1.9 bracnh

gradle/wrapper/gradle-wrapper.properties
RlonRyan: Not if you update forge properly, it's only overwritten by the gradle wrapper task.

build.gradle
RlonRyan: They can be undone (i.e. I can undo them for you), but keeping to an aggressive trailing white space removal policy can really help prevent future merge conflicts.

.travis.yml
RlonRyan: I should have removed that right off the bat, but as a suggestion, I can configure it to match any setup you want, if you just tell me how you would like it.

.travis.yml
RlonRyan: This prevents the cache from being constantly repackaged, in my experience, with the most possible amount of files preserved in the cache.

RlonRyan:
So... What needs to be done here?

PR #3674, opened 2016-09-08
Update zh_CN.lang

PeakXing:
Delete unused fields
Add new fields
Modify some fields


HenryLoenwind:
complete nonsense for 1.7.10

PR #3683, opened 2016-09-11
Solar panels

vacallard:
hello. i love ender io and i use it a lot when im starting a new world. and for early power gen, solar panels are awesome. but in ender io they take to much effort and resources for whats not a great power generation. i think that either the recipie or the amount of power it produces should be balenced. tanks for your time.


MatthiasMann:
WTF did you do? why did you create a pull request instead of an issue report ?!?

PR #3690, opened 2016-09-13
Fix Fluid Conduit failure to extract

codewarrior0:
A Fluid Conduit adjacent to multiple fluid containers, set to "Extract" on all of them, will stop extracting as soon as any of them become empty. A quick look at the source code suggests this line is the problem. 

So, I'm giving you a completely untested change to a no-longer-maintained branch of your project just in case somebody finds it useful. Cheers!

### DIFF
--- a/src/main/java/crazypants/enderio/conduit/liquid/LiquidConduit.java
+++ b/src/main/java/crazypants/enderio/conduit/liquid/LiquidConduit.java
@@ -145,7 +145,7 @@ public class LiquidConduit extends AbstractTankConduit {
             }
           }
           if(!foundFluid) {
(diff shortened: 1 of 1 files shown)


HenryLoenwind:
Is this only for Ender IO 2.2 or would this also apply to 2.3? (branch master)

codewarrior0:
Looks like it can be applied to master: https://github.com/SleepyTrousers/EnderIO/blob/master/src/main/java/crazypants/enderio/conduit/liquid/LiquidConduit.java#L143

codewarrior0:
#3691 is this same change, based on master.

PR #3843, opened 2016-11-17
Fix Immersive's hemp haversting

Prototik:
Actually EnderIO's farmer havresting Immersive Engineering's Industrial Hemp wrong: it harvest base block too (in addition to upper block). It causes re-plant loop and additional waiting time.

How it works now:
screenshot 1

How it should works:
screenshot 2
screenshot 3


HenryLoenwind:
I think this was done on purpose to harvest the seeds. Those are more interesting than the hemp as they are used to produce seed oil which goes into power generation.

Prototik:
@HenryLoenwind haversting upper block still drop seeds and hemp's fiber itself. But with this patch my farm havrest ~3 times faster.

HenryLoenwind:
thx for checking

PR #3891, opened 2016-12-07
Fix for Applied Energistics 2 SAG Mill Recipes

tgstyle:
Changed to ore dictionary entries where available, as well as new item ids.


HenryLoenwind:
is this really for Minecraft 1.7.10?

tgstyle:
My bad, it is for 1.10.2. I will re-submit it.

HenryLoenwind:
ok, thx.

PR #3936, opened 2016-12-21
Rewrite LootManager to use a standard Loot Table. Closes #3927

### LINKED ISSUE #3927: EnderIO causes GSON to throw an exception when certain loot tables are serialised
#### Issue Description:

When LootTweaker attempts to serialise certain loot tables(to dump them to a folder), GSON throws an error. This error is not caused by LootTweaker or CraftTweaker, as it only occurs with EnderIO installed. EnderCore is not the cause of the problem. LootTweaker successfully dumps loot tables modified by other mods, such as Actually Additions and BetterBeginnings.

#### What happens:

When LootTweaker attempts to serialise certain loot tables with EnderIO installed, GSON throws an error.

LootTweaker to successfully dump all loot tables.

#### Steps to reproduce:

1. Install [this build](https://drive.google.com/open?id=0B_IqTUgN98IXdWNKNHRtOF9UblU) of LootTweaker, also install CraftTweaker(I reproduced this issue with CT 1.10.2-3.0.15, but any 1.10.2 build should do). Any build of LootTweaker will reveal the issue, but the linked one has been tweaked to give a more useful error.
2. Launch Minecraft and open a world. 
3. Run the command /mt loottables all. There will be no errors in the log
4. Install EnderIO and EnderCore.
5. Repeat step 3. Errors will now occur.

____

- CraftTweaker: 1.10.2-3.0.15
- EnderIO: 1.10.2-3.0.1.142_beta
- EnderCore: 1.10.2-0.4.1.61-beta
- Minecraft: 1.10.2
- Forge: 1.10.2-12.18.2.2134

https://paste.ee/p/TNh8M

Leviathan143:
See title. This fixes the issue discussed in #3927 
Loot weights will need to be adjusted as the loot weight system behaves differently to the previous probability based system. I tried to do this myself mathematically and failed due to differences between the two systems. The values will have to be adjusted by a process of guess and check, or something better if anyone knows of it.

A test world that can be used to easily test the loot tables can be found [here](https://drive.google.com/open?id=0B_IqTUgN98IXaVdjV0JtcFphZU0).
The world contains 4 buttons that each place a chest that uses one of the four loot tables modified by EnderIO.

### DIFF
--- a/src/main/java/crazypants/enderio/LootManager.java
+++ b/src/main/java/crazypants/enderio/LootManager.java
@@ -65,82 +65,105 @@ public class LootManager {
 
     if (evt.getName().equals(LootTableList.CHESTS_SIMPLE_DUNGEON)) {
 
-        lp.addEntry(createLootEntry(itemAlloy.getItem(), Alloy.DARK_STEEL.ordinal(), 1, 3, 25, Config.lootDarkSteel));
-        lp.addEntry(createLootEntry(itemConduitProbe.getItem(), 10, Config.lootItemConduitProbe));
-        lp.addEntry(createLootEntry(Items.QUARTZ, 3, 16, 25, Config.lootQuartz));
-        lp.addEntry(createLootEntry(Items.NETHER_WART, 1, 4, 20, Config.lootNetherWart));
-        lp.addEntry(createLootEntry(Items.ENDER_PEARL, 1, 2, 30, Config.lootEnderPearl));
-        lp.addEntry(createLootEntry(DarkSteelItems.itemDarkSteelSword, 10, Config.lootTheEnder));
-        lp.addEntry(createLootEntry(DarkSteelItems.itemDarkSteelBoots, 10, Config.lootDarkSteelBoots));
-        lp.addEntry(createLootCapacitor(15, true));
-        lp.addEntry(createLootCapacitor(15, true));
-        lp.addEntry(createLootCapacitor(15, true));
-        //175
-    } else if (evt.getName().equals(LootTableList.CHESTS_VILLAGE_BLACKSMITH)) {
-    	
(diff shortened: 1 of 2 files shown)


src/main/java/crazypants/enderio/LootManager.java
```diff
@@ -72,181 +61,99 @@ public void onLootTableLoad(LootTableLoadEvent evt) {
...
 
-  private LootItem createLootEntry(Item item, int minSize, int maxSize, double chance) {
-    return new LootItem(new ItemStack(item), chance, minSize, maxSize);
+  private LootEntry createLootEntry(Item item, int minSize, int maxSize, int chance, boolean enabled) {
+    return createLootEntry(item, 0, minSize, maxSize, chance, enabled);
   }
 
-  private LootItem createLootEntry(Item item, int ordinal, int minStackSize, int maxStackSize, double chance) {
-    return new LootItem(new ItemStack(item, 1, ordinal), chance, minStackSize, maxStackSize);
+  /**If enabled is true, an empty loot entry of the same weight is added.
```
HenryLoenwind: "If enabled is false"

HenryLoenwind:
I'll need to look at this in my dev env later, it's a bit too large to really see in the diff view on github.

Leviathan143:
After talking abou this on the MMD Discord with several people, MysteryM pointed out that I could use the RandomChance LootCondition instead of weights. I have done this, this implementation should behave similarly to the old system. It will also not require tweaking of loot values, though checking the loot spawn rates are still as intended is still a good idea. Empty loot entries are also no longer needed to preserve the loot spawn rates when loot is turned on/off in the config.

HenryLoenwind:
nice.

PR #3988, opened 2017-01-13
Modified Molten Ender Teleport

KetHalpak:
Modified the teleportation for molten ender to trigger an event as if an ender pearl has been thrown for the teleportation of a mob/player. I did this mainly to work with the Dark Utils Ender Tether as it adjusts the teleport destination based on the EnderTeleportEvent. Teleport still behaves as normal and random if there is nothing that intercepts the event.

### DIFF
--- a/src/main/java/crazypants/enderio/fluid/BlockFluidEio.java
+++ b/src/main/java/crazypants/enderio/fluid/BlockFluidEio.java
@@ -51,6 +51,7 @@ import net.minecraftforge.client.event.RenderBlockOverlayEvent;
 import net.minecraftforge.client.event.RenderBlockOverlayEvent.OverlayType;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.entity.living.LivingSpawnEvent;
+import net.minecraftforge.event.entity.living.EnderTeleportEvent;
 import net.minecraftforge.fluids.BlockFluidClassic;
 import net.minecraftforge.fluids.Fluid;
 import net.minecraftforge.fluids.IFluidBlock;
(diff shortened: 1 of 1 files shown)


HenryLoenwind:
Your formatting is off, can you please fix that?

Also, please fire a TeleportEntityEvent, that's the one for entities being teleported by something else. The EnderTeleportEvent is for entities that teleport themselves. You can fire that one for Endermen if you want, so it doesn't make a difference if they teleport by themselves because WATER!!!! or are being teleported by it.

KetHalpak:
I've used EnderTeleportEvent as the motivation behind this adjustment is to have the teleport function as if it were an ender pearl throw by an entity. This allows existing mods that provide control using EnderTeleportEvent such as Botania and Dark Utils to control the destination of mobs that fall into the fluid. Using TeleportEntityEvent would result in the desired outcome not being achieved for this adjustment.

HenryLoenwind:
Ok, makes sense. but then we should do the whole Ender Pearl spiel, including Endermite chance, dismounting (also check the other way around, e.g. teleporting a horse should also dismount its rider) and fall damage (reduce that to 1 for mobs with less than 10 max health and 0 for non-living entities---we don't want to kill them outright)...

thiakil:
NB at EIO version 1.10.2-0.4.1.62-beta at least its broken-ish if you step into the fluid while mounted on a horse/donkey

KetHalpak:
That is going beyond what I wanted to achieve here, just had an idea on how to make this interesting/mildly useful in a way that isn't overpowered so thought I'd give it a go myself rather than just saying 'hey can you do this?'. If you want to keep the teleportation as purely random I'm not bothered.

HenryLoenwind:
Ok, I'll use this and base and add the rest.

PR #4050, opened 2017-02-21
Update ko_KR.lang

01QueN10:


01QueN10:
OK, I'll update this soon.

PR #4103, opened 2017-03-17
Fixed inconsistent punctuation

PapiDimmi:


vippeople22 (approved):
i mean it seems just fine perhaps someone just added a period out of habit?

PR #4160, opened 2017-04-16
Implement vanilla TileEntity rotation/mirroring functions

fnuecke:
This implements the vanilla rotate/mirror methods in TileEntities that are rotatable. This allows mods that want to rotate blocks to do so without additional special code.


HenryLoenwind:
Sorry, this got lost on accident

Issue #4096, opened 2017-03-14
Item Conduit can duplicate items with AE2


xsun2001:
[xsun2001/Applied-Energistics-2-Unofficial#2](https://github.com/xsun2001/Applied-Energistics-2-Unofficial/issues/2)
Although 1.7.10's enderio may not be in the maintenance, but it is only the enderio items conduit will be such a problem. Although I continue to maintain the 1.7.10's AE2, but I do not understand the way you work in the conduit , so I can not fix this problem in AE2. Would you please tell me how the enderio network works or how it triggers the network refresh. Thank you!!!

xsun2001:
No.In fact, there is only AE2 and EnderIO.

HenryLoenwind:
I would need to look into the the code, but from memory: Conduits do their work in the server tick event. They keep a list of neighbor tile entities and refresh it when they get a Block.neighborChanged().

xsun2001:
AE2's Annihilation Plane only uses
`w.setBlock( x, y, z, Platform.AIR_BLOCK, 0, 3 );`
to break a block.
Can it fire the enderio network refresh?

xsun2001:
It seems that you didn't add the `onNeighborChanged` method to ItemConduit in 1.7.10.

> ![9k 0fl g8p7cvn y6 c9jn](https://cloud.githubusercontent.com/assets/14010993/25551383/a7e1e240-2cb6-11e7-8199-761d46f8b129.png)
left is `master` branch which is for mc1.7.10 and right is `1.10` branch.

Now I want to get you premission to change your bytecode at runtime by ASM to add that method. Is there anything to notice to avoid breaking the CC0 1.0 and something annoying?

xsun2001:
@HenryLoenwind Have you seen that??

HenryLoenwind:
Interesting. I wasn't aware of this.

Please don't ASM this in, send me a PR instead. Our build pipeline for 1.7.10 still exists, so I can easily push out a new release.

PR #4212, opened 2017-05-10
Add onNeighborChange method to ItemConduit to deal with a duplicate items bug with AE2

xsun2001:
#4096
https://github.com/xsun2001/Applied-Energistics-2-Unofficial/issues/2


xsun2001:
@HenryLoenwind I have added the onNeighborChange method to ItemConduit but it doesn't solve this problem. I'd like to know what I have to do but I forgot to do. I really want to solve this bug smoothly.

PR #4256, opened 2017-06-13
:Pl 'l554trgfhtfbg ffhbf gh

Bossman195:
wfsef ds fdf erd fcv vfdvfd ds


tterrag1098:
The "monkey's on typewriters" was a thought experiment, you weren't supposed to actually do it.

PR #4401, opened 2017-12-31
Fixed pickaxe interrupting while charging

MatanShahar:
Fixed the issue where the pickaxe (or any other tool) will be interrupted while it's being charged by a wireless charger.
The issue was happening because the default behavior was to reset block break progress on and `ItemStack` change, that includes NBT which is used to store the tools' energy.

The current implementation tries to as strict as possible when it comes to resetting the progress. This is done by running a deep equality comparison of the old and new NBT trees while ignoring the `energy` key.

I'm not a big fan of the current state of things though, feel free to suggest a different approach.
What I had in mind was:
* Store the energy value on an internal variable and to (re)store it from/to NBT when needed. Not sure how to do this. And client-server sync may be a problem.
* Separate the upgrade entirely from the pick and use some sort of compose-able item proxy

### DIFF
--- a/src/main/java/crazypants/enderio/base/item/darksteel/ItemDarkSteelPickaxe.java
+++ b/src/main/java/crazypants/enderio/base/item/darksteel/ItemDarkSteelPickaxe.java
@@ -411,4 +411,32 @@ public class ItemDarkSteelPickaxe extends ItemPickaxe implements IAdvancedToolti
     }
   }
 
+  @Override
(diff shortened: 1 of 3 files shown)


HenryLoenwind:
I don't think we actually need this level of aggressiveness. It's not as if the user could change their stack in the middle of breaking a block in any easy way---they'd need a server-side mod to do this for them.

PR #4423, opened 2018-01-11
Telepad Models

Mr-Pyro:
Split the telepad model into center, side, corner, and crossbeams.


HenryLoenwind:
It will take me a couple of days to get to this. Thx!

PR #4439, opened 2018-01-23
Conduits Now Compiles

EpicSquid:
Way too much to document, sorry for all bugs.


enderio-base/src/main/java/crazypants/enderio/base/filter/gui/PowerItemFilterGui.java
```diff
@@ -108,7 +101,8 @@ public void actionPerformed(GuiButton guiButton) {
 
   private void sendFilterChange() {
     updateButtons();
-    PacketHandler.INSTANCE.sendToServer(new PacketItemConduitFilter(itemConduit, gui.getDir()));
+    filterContainer.onFilterChanged();
+    //PacketHandler.INSTANCE.sendToServer(new PacketItemConduitFilter(itemConduit, gui.getDir()));
```
Ricket: Is there a repercussion to not doing `PacketHandler.INSTANCE.sendToServer`? For example, does this now only work in single player mode?

HenryLoenwind:
Manually merged

enderio-base/src/main/java/crazypants/enderio/base/filter/gui/PowerItemFilterGui.java
HenryLoenwind: The call was moved to the container to decouple the GUI.

PR #4448, opened 2018-02-06
Small Conduit Fixes

EpicSquid:
But without Stuffing Up the Import Order


enderio-conduits/src/main/java/crazypants/enderio/conduit/liquid/AbstractLiquidConduit.java
```diff
@@ -73,7 +73,7 @@ public void setExtractionRedstoneMode(@Nonnull RedstoneControlMode mode, @Nonnul
   public RedstoneControlMode getExtractionRedstoneMode(@Nonnull EnumFacing dir) {
     RedstoneControlMode res = extractionModes.get(dir);
     if (res == null) {
-      res = RedstoneControlMode.NEVER;
```
HenryLoenwind: This is on purpose so they don't fill up with junk while placing them (conduits that have content don't auto-connect anymore).

PR #4453, opened 2018-02-12
Some work on Filters

EpicSquid:
Currently only "works" on ItemFilters. NBT writing seems to work but reading does not.


enderio-base/src/main/java/crazypants/enderio/base/filter/items/ItemBasicItemFilter.java
```diff
@@ -92,4 +105,33 @@ public void addInformation(@Nonnull ItemStack stack, @Nullable World worldIn, @N
     }
   }
 
+  @Override
+  @Nonnull
+  public ActionResult<ItemStack> onItemRightClick(@Nonnull World worldIn, @Nonnull EntityPlayer playerIn, @Nonnull EnumHand handIn) {
+    if (playerIn.isSneaking() && ModObjectRegistry.getModObjectNN(this).openGui(worldIn, playerIn.getPosition(), playerIn)) {
```
HenryLoenwind: you may want to put the hand into openGui instead of the player position (the blockpos parameter is optional, I wouldn't put in data that is not needed just because the param is there) instead of hardcoding mainHand. Afterall, this method will be called for off-hand istemstacks, too.

HenryLoenwind:
Also, please don't use GUIs textures with optionally painted slot backgrounds. It doesn't play nice with ghost slots.

PR #4455, opened 2018-02-13
Revert hack to fix McJtyLib crashing us

josephcsible:
I fixed McJtyLib to not bundle our API, so upgrading to McJtyLib 2.5.2 makes this no longer necessary.


josephcsible:
Can this be merged, so that it becomes practical for other mods to use this API?

EpicSquid:
I will make a quick check to fix this later, there is one thing I've changed since then that may effect this. Rest assured this will be merged by the end of tonight @josephcsible

PR #4481, opened 2018-03-22
Enchanter COFH, Draconic Evolution, EnderZoo Support

tyler489:


EpicSquid:
If we are to add this it would probably be best to make a new recipe file for modded enchanter recipes, similar to the sagmill.

tyler489:
If you wish.

HenryLoenwind:
split of not doesn't matter to me, we can move recipes around now without any worries. But I'm not ok with the itemSkull change. First you now got a conflict between those 2 recipes, and then we should not change old recipes without very good reason. People know them and expect them to stay the same.

tyler489:
No because tormented endermand head doesnt extend itemskull in game. But i can change it back

tyler489:
Ill fix vorpal when i get home. Any ideas on  what to use?

tyler489:
Obsid will work.

PR #4482, opened 2018-03-22
Fixes ae"Compatibility" disabling all wooden gear

tyler489:
Ae2's gear is still registered even when disabled and the recipy conflicts. when it is disabled it willnot work with enderio machines and the ae2 dev wont fix.


EpicSquid:
There is no need for this, you can disable the option in the config and Ae2's end

HenryLoenwind:
Nope. This is the standard wooden gear recipe used by all mods that have wooden gears.

PR #4483, opened 2018-03-23
Sould Binder Fix

Mr-Pyro:
Temperary fix for the soul binder (can now only craft if the output slot is empty (prevents soul vial (bat) + broken spawner from creating a slime spawner if a slime spawner is already in that slot)).  canStartNextTask (should) check if the recipe output is equal to the item in the output slot or if the slot is empty.


HenryLoenwind:
oh, so that's the issue.

Then this test is wrong: https://github.com/Mr-Pyro/EnderIO/blob/83f35a964762b2fd797f0406ac70379312a85dad/enderio-base/src/main/java/crazypants/enderio/base/machine/baselegacy/AbstractPoweredTaskEntity.java#L341

PR #4487, opened 2018-03-24
Redstone conduit and Dialer bug fixes

EpicSquid:


enderio-machines/src/main/java/crazypants/enderio/machines/machine/teleport/telepad/TileDialingDevice.java
```diff
@@ -82,7 +82,11 @@ public void removeTarget(TelepadTarget target) {
     if (target == null) {
       return;
     }
-    targets.remove(target);
+    for (TelepadTarget t : targets) {
+      if (target.equals(t)) {
+        targets.remove(t);
```
HenryLoenwind: * CME?
* Isn't that exactly what remove() does? Removing on equals()...

PR #4493, opened 2018-03-28
Fixed Baubles, added Advanced Big Item Filter and Fixed Enchanter Recipes

EpicSquid:


enderio-base/src/main/java/crazypants/enderio/util/ShadowInventory.java
```diff
@@ -25,8 +26,8 @@ public int getSizeInventory() {
 
   @Override
   public @Nonnull ItemStack getStackInSlot(int index) {
-    final ItemStack itemStack = items[index];
-    return itemStack == null ? master.getStackInSlot(index) : itemStack;
+    final ItemStack itemStack = items.get(index);
+    return itemStack.isEmpty() ? master.getStackInSlot(index) : itemStack;
```
HenryLoenwind: This actually is a logic error. It must be possible to have an empty itemstack in the shadow inventory override an existing one in the shadowed one...

EpicSquid: Except ShadowInventory isn't even used any more because they fixed the client-server sync. Why did I keep it.

PR #4528, opened 2018-04-04
Fixed Overlapping Buttons

Mr-Pyro:
Moved the visible button in the travel accessible gui, so that it no longer overlaps with the button that switches the gui.

### DIFF
--- a/enderio-machines/src/main/java/crazypants/enderio/machines/lang/Lang.java
+++ b/enderio-machines/src/main/java/crazypants/enderio/machines/lang/Lang.java
@@ -91,8 +91,6 @@ public enum Lang implements ILang {
   GUI_AUTH_ERROR_HARVEST(".chat.travel_accessable.error.harvest"),
   STATUS_TELEPAD_UNFORMED(".status.telepad.unformed"),
   GUI_TELEPAD_NOFLUID(".chat.telepad.noFluid"),
-  GUI_TELEPAD_CLOSE(".gui.telepad.close"),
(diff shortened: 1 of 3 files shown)


enderio-machines/src/main/java/crazypants/enderio/machines/lang/Lang.java
```diff
@@ -91,6 +91,8 @@
   GUI_AUTH_ERROR_HARVEST(".chat.travel_accessable.error.harvest"),
   STATUS_TELEPAD_UNFORMED(".status.telepad.unformed"),
   GUI_TELEPAD_NOFLUID(".chat.telepad.noFluid"),
+  GUI_TELEPAD_CLOSE(".gui.telepad.close"),
+  GUI_TELEPAD_CLOSE_2(".gui.telepad.clsoe"),
```
EpicSquid: Spelling error, so this won't work. Also what are these for again?

enderio-machines/src/main/java/crazypants/enderio/machines/machine/teleport/GuiTravelAccessable.java
```diff
@@ -29,11 +30,15 @@
   private static final int ID_PUBLIC = 0;
   private static final int ID_PRIVATE = 1;
   private static final int ID_PROTECTED = 2;
+  
+  //private static final int ID_CLOSE_WINDOW_BUTTON = 12615;
```
EpicSquid: Can we remove references to the close window button for now? More than likely it will need a full rewrite to implement, so this commented code may not be relevant ultimately.

PR #4532, opened 2018-04-05
add canInsertItem check before do insert

q178666380:
machine implements iSidedInventory should invoke canInsertItem method to  ensure item can insert into


HenryLoenwind:
have you even looked at doInsertItem()?

q178666380:
this method only check isMachineItemValidForSlot(),but such as TileTransceiver also overwrite
the method canInsertItem ,TileTransceiver.canInsertItem() look like it donot want to put same item to diff slot,but TileTransceiver work not so when in game on insert item into it.

PR #4540, opened 2018-04-07
Potentialy fix incompatibily with diet hoppers in 1.12

tyler489:
Rwtema said to use Blocks.Hopper or  forge "IRegistryDelegate"


tyler489:
Even if this doesnt fix it it will make the code match the item filters using blockHopper

HenryLoenwind:
We are using `Blocks.HOPPER`. If @rwtema breaks this one, they are really doing something wrong.

tyler489:
We call blockhopper for item filter but minecraft:hopper for the impulse hopper.

HenryLoenwind:
not a fix

PR #4542, opened 2018-04-08
Beat thermal fluid ores with nerf bat

tyler489:


EpicSquid:
Why the nerf?

tyler489:
produced WAY more than thermal did even without Grinding balls

PR #4550, opened 2018-04-13
Fixed TravelController Not Enough Power Message

Mr-Pyro:
"Not Enough Power" message is now only displayed once per right click.


EpicSquid:
This is quite a hack, perhaps try using tt's idea and check the hand and world.isRemote in the item instead? Since it is only effecting the staff of travelling

enderio-base/src/main/java/crazypants/enderio/base/item/travelstaff/ItemTravelStaff.java
```diff
@@ -61,7 +61,7 @@ public ICapabilityProvider initCapabilities(@Nonnull ItemStack stack, @Nullable
 
   @Override
   public @Nonnull ActionResult<ItemStack> onItemRightClick(@Nonnull World world, @Nonnull EntityPlayer player, @Nonnull EnumHand hand) {
-    ItemStack equipped = player.getHeldItem(hand);
+	  ItemStack equipped = player.getHeldItem(hand);
```
EpicSquid: Can we undo this?

PR #4575, opened 2018-04-17
Re-add ComputerCraft bundled cable integration

SquidDev:
Effectively cherry-picked from #1208. I'm not sure if this is a) wanted and b) how mod integration is meant to be handled, so feel free to close it :).

One thing worth noting is that `IRedstoneConduit.getNetworkOutputs` has the side parameter annotated as `@Nonnull`, but handles `null` arguments (which is required here in order to return the signal for all colours).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -95,6 +95,7 @@ allprojects {
         maven { url = "http://mod-buildcraft.com/maven/" }
         maven { url = "https://maven.latmod.com" } // FTBLib
         maven { url = "http://maven.covers1624.net/" } // CoFHCore
+        maven { url = "http://cc.crzd.me/maven/" } // ComputerCraft
         ivy {
             url "http://ae-mod.info/builds"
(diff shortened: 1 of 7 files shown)


EpicSquid:
I definitely have been wanting to add this! Thank you. I will check it over properly later. Main thing to note, which computer craft fork at this point will this be officially supporting?

SquidDev:
> Main thing to note, which computer craft fork at this point will this be officially supporting?

It should support all of them, as the API hasn't diverged (yet!). It's currently just building off the master branch of the original repo, as I'd consider that the a good mix of official and stable, though I'm happy to switch it to use the latest Curse release instead.

EpicSquid:
No need to change it, I'll look into this soon

SquidDev:
Thanks!

EpicSquid:
No worries

PR #4625, opened 2018-05-13
Hoe upgrade stuff

LumberWizard:
Pretty much that.


HenryLoenwind:
Please use the new config system, the old config.java is on its way out.

enderio-base/src/main/java/crazypants/enderio/base/item/darksteel/upgrade/hoe/HoeUpgrade.java
```diff
@@ -16,7 +16,7 @@
   public static final @Nonnull HoeUpgrade INSTANCE = new HoeUpgrade();
 
   public HoeUpgrade() {
-    super(UPGRADE_NAME, "enderio.darksteel.upgrade.hoe", new ItemStack(Items.DIAMOND_HOE), Config.darkSteelHoeCost);
+    super(UPGRADE_NAME, "enderio.darksteel.upgrade.hoe", new ItemStack(Items.DIAMOND_HOE), DarkSteelConfig.darkSteelHoeCost.get());
```
HenryLoenwind: use the `super(String, String, ItemStack, IValue<Integer>)` here, not the `super(String, String, ItemStack, int)` one

PR #4639, opened 2018-05-16
Fix typo in Enderium Base recipe

codewarrior0:


tyler489:
this could actually be simplified even more by just doing ```      <input name="enderpearl" amount="4"/>```

PR #4640, opened 2018-05-17
Add a JEI Category for Grinding Balls

codewarrior0:
Lists all known Grinding Balls alongside their stats. Pointing at each
stat shows a short tooltip explaining its effects.

The category can be found by checking the usage for the Sag Mill, or
by checking the usage or recipes for any grinding ball.

image


codewarrior0:
ballgrinding

PR #4658, opened 2018-05-24
Fixes IEnergyStorage implementation generating infinite rf

SonarSonic:
See [Forge Documentation](https://github.com/MinecraftForge/MinecraftForge/blob/1.12.x/src/main/java/net/minecraftforge/energy/IEnergyStorage.java#L51)

The `extractEnergy` method in `PowerConduit` should return the power actually extracted as seen in the documentation above.

At the moment it returns the max input/output of the conduit instead.
Any other mods which attempt to draw power from a nearby cable will therefore create an infinite energy source as the block will assume the max input/output was extracted.

I've changed the method to use the internal energy storage of the conduit like the `receiveEnergy` method, but you may wish to implement this another way as I haven't looked into the deeper operation of the conduit power network and really I just wanted to bring this to your attention.

### DIFF
--- a/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/power/PowerConduit.java
+++ b/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/power/PowerConduit.java
@@ -366,7 +366,11 @@ public class PowerConduit extends AbstractConduit implements IPowerConduit, ICon
     if (getMaxEnergyIO(subtype) == 0 || maxExtract <= 0) {
       return 0;
     }
-    return getMaxEnergyIO(subtype);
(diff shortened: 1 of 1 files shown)


HenryLoenwind:
That actually is a porting bug. extract should always return 0, as Ender IO only supports the push-based energy model.

(see https://github.com/SleepyTrousers/EnderIO/blob/release/1.7.10/2.3/src/main/java/crazypants/enderio/conduit/power/PowerConduit.java#L392 )

SonarSonic:
I thought so, is it possible to have any control over the external extraction of energy? I assume you keep it all internal to maintain an even distribution of energy?

EpicSquid:
Fixed as of d9892f9

PR #4749, opened 2018-07-06
Fix 4565

tyler489:
Fix #4565 being in the wrong file..


tyler489:
compressed into #4751

PR #4752, opened 2018-07-06
End Steel gear can be upgraded to Empowered V

thundR:
Couple things:
1. Currently the power/damage ratio is 95 percent. Should it be pushed to 100 to allow players to eventually unlock fully powered gear?
2. The means for asserting tier is really ugly and I'm open to ideas for better solutions. We could also just merge this and you guys can work on a full IDarkSteelItem refactor down the road. 
3. This implicitly adds another tier of the related empowered properties for things like swords. 
4. Recipe is an ender crystal + 15 levels. Very open to changes.


EpicSquid:
Can we make the tier an Enum instead of an integer?

thundR:
Could use another look I think

HenryLoenwind:
What about having getEquipmentData() instead of getTier()? That way the upgrades would have more control about what they can be applied to.

PR #4781, opened 2018-07-17
remove custom aspects to restore Thaumcraft functionality

jdratlif:
This change just removes the loadAspects call to restore thaumcraft aspects that are broken by the integration until this problem is resolved.

### DIFF
--- a/enderio-base/src/main/java/crazypants/enderio/base/integration/thaumcraft/ThaumcraftUtil.java
+++ b/enderio-base/src/main/java/crazypants/enderio/base/integration/thaumcraft/ThaumcraftUtil.java
@@ -16,7 +16,7 @@ public class ThaumcraftUtil {
 
   public static void create() {
     if (Loader.isModLoaded("thaumcraft")) {
-      ThaumcraftAspects.loadAspects();
+      //ThaumcraftAspects.loadAspects();
     }
   }
 


EpicSquid:
I have tested the thaumcraft integration bug and it does not remove aspects for me. Simply removing the integration is not a sufficient fix.

jdratlif:
Here is my test.

Create new profile in Twitch MC 1.12.2 Forge 2705. Add Ender IO (5.0.29) and Thaumcraft (beta 17).
In a new world, coal and coal blocks have no aspects. It's not just a graphical glitch. When you toss them in crucible, you get nothing.

In my main mod pack, I also lose Gelum on snowball and snow. This does not happen in my small test. It may be another mod interfering or some odd interaction.

Are you saying that you have aspects on coal and coal blocks when using Ender IO 5.0.29 (Ender Core 0.5.35)?

I have tested another patch that changes the line from adding coal dust aspects by getting aspects on coal to adding the specific aspects. That works, but it doesn't fix Gelum on snow/snowballs in my main pack, where commenting out the loadAspects() line does.

There may be other things that lose aspects, but these are the ones I've seen so far.

jdratlif:
https://github.com/SleepyTrousers/EnderIO/pull/4787

How about this fix instead?

HenryLoenwind:
fixed with cb00da0c1b79dfb34646848a7ee5db1171e7ac08

PR #4785, opened 2018-07-19
ADD Nuclearcraft alloys

tyler489:


enderio-base/src/main/resources/assets/enderio/config/recipes/base.xml
```diff
@@ -693,7 +693,87 @@ XML editor will display that as tooltips when editing this file.
...
+    <alloying energy="4000" exp="0">
+      <input name="ingotSteel"/>
+      <input name="ingotBoron"/>
+      <output name="ingotFerroboron" amount="2"/>
+    </alloying>
+  </recipe>
+ 
+  <recipe name="Alloy: FerroBoron (Dusts) (Nuclearcraft)" required="false">
+    <alloying energy="4000" exp="0">
+      <input name="dustSteel"/>
```
HenryLoenwind: We don't do dust recipes in addition to ingot alloying recipes because that interferes with the dust->ingot smelting.

PR #4902, opened 2018-10-17
Added padding upgrade config localization

LumberWizard:


tyler489:
Request changes.. lang must say Henry's Headbang protection

LumberWizard:
That is not what the upgrade does unfortunately... It's a sound muffler, look in 3a2ddf1d37e77735a455cc9f4167d00034f69ba8.

PR #4933, opened 2018-11-02
Added fluids translation

DoctorFTB:


LumberWizard:
Should be "Molten <metal name>"

HenryLoenwind:
thanks, I forgot about those. Can you also add the block names while you're at it?

DoctorFTB:
About "tile.fluid.energetic_silver.name" - I not found where this.

pupnewfster:
I think Henry is saying it should be added to the lang file and doesn't exist at the moment. Similar to how https://github.com/SleepyTrousers/EnderIO/blob/master/enderio-integration-tic/src/main/resources/assets/enderio/lang/en_us.lang#L39 it exists there for Electrical steel

HenryLoenwind:
oops, sorry, I forgot about this not being merged.

I added the block names myself and also moved the names to tic-integration, as that is the module that actually adds the fluids.

PR #4964, opened 2018-11-27
Allow conduits to decide what items they can accept as filters/upgrades

pupnewfster:
Implements the TODO about isFilterUpgradeAccepted in InventoryUpgrades
I ran into some issues while attempting to make a custom filter for a new set of conduits I am implementing and tracked it down to [this](https://github.com/SleepyTrousers/EnderIO/blob/master/enderio-conduits/src/main/java/crazypants/enderio/conduits/gui/InventoryUpgrades.java#L132) in InventoryUpgrades.java
```java
// TODO make a way to do this based on the conduit
private boolean isFilterUpgradeAccepted(@Nonnull ItemStack stack, IConduit con, boolean isInput) {
```

I decided to implement the TODO so that custom conduits can decide for themselves if they can accept a stack as a filter. While implementing this change I also ran into a few bugs where filters/upgrades could be voided in some conduits by shift clicking them in. For example, Redstone Conduits voiding speed upgrades, and Pressurized Liquid Conduits voiding fluid filters. This pull has the side effect of also fixing those bugs.

This pull changes InventoryUpgrades so that only classes that implement IFilterHolder can accept filters, and only classes that accept IUpgradeHolder can accept upgrades.

How it does this:
- IFilterHolder has a new default method `boolean isFilterUpgradeAccepted(@Nonnull ItemStack stack, boolean isInput)`, that allows Filter Holders to specify if they consider an ItemStack a valid filter. This method defaults to false.
- IUpgradeHolder has a new default method `boolean isFunctionUpgradeAccepted(@Nonnull ItemStack stack)`, that defaults to true. This means that any conduit that also is an Upgrade Holder can accept any upgrades, but also allows for them to be more restrictive if they choose to check additional information about the ItemFunctionUpgrade. Before this method is called InventoryUpgrades ensures that the stack is an ItemFunctionUpgrade so as to ensure that the stack is some form of Function Upgrade.

### DIFF
--- a/enderio-conduits/src/main/java/crazypants/enderio/conduits/gui/InventoryUpgrades.java
+++ b/enderio-conduits/src/main/java/crazypants/enderio/conduits/gui/InventoryUpgrades.java
@@ -6,17 +6,7 @@ import crazypants.enderio.base.conduit.IConduit;
 import crazypants.enderio.base.conduit.item.ItemFunctionUpgrade;
 import crazypants.enderio.base.filter.IFilter;
 import crazypants.enderio.base.filter.capability.IFilterHolder;
-import crazypants.enderio.base.filter.fluid.items.IItemFilterFluidUpgrade;
-import crazypants.enderio.base.filter.item.items.IItemFilterItemUpgrade;
-import crazypants.enderio.base.filter.item.items.ItemBasicItemFilter;
-import crazypants.enderio.base.filter.redstone.items.IItemInputSignalFilterUpgrade;
-import crazypants.enderio.base.filter.redstone.items.IItemOutputSignalFilterUpgrade;
 import crazypants.enderio.conduits.capability.IUpgradeHolder;
-import crazypants.enderio.conduits.conduit.item.IItemConduit;
-import crazypants.enderio.conduits.conduit.liquid.AbstractTankConduit;
-import crazypants.enderio.conduits.conduit.liquid.EnderLiquidConduit;
-import crazypants.enderio.conduits.conduit.power.IPowerConduit;
-import crazypants.enderio.conduits.conduit.redstone.IRedstoneConduit;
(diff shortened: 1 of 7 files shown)


EpicSquid:
This looks fine at a glance. I'll test it more later today when I'm at home. What conduits asre you working on out of curiosity?

enderio-conduits/src/main/java/crazypants/enderio/conduits/gui/InventoryUpgrades.java
```diff
@@ -129,25 +122,11 @@ public boolean isItemValidForSlot(int slot, @Nonnull ItemStack stack, IConduit c
     return false;
   }
 
-  // TODO make a way to do this based on the conduit
   private boolean isFilterUpgradeAccepted(@Nonnull ItemStack stack, IConduit con, boolean isInput) {
```
HenryLoenwind: Please make both either methods or inline in the switch. I don't care which one, but they should be handled the same way.

pupnewfster:
Mekanism gas conduits (as I know from reading that you guys don't want to deal with all the issues that would come from including them). I am putting/will put messages all over the place basically saying to report the issues to me rather than to either EnderIO or Mekanism directly. But it is just so nice being able to have two paths of wires/cables cross for compact machine setups that I decided to just do it myself in its own addon.

EpicSquid:
Sounds good. I was guessing that. Give me a buzz if you need any help with it. It should be pretty straightforward. My personal preference would be for them to match the fluid conduit tiering but it's up to you.

pupnewfster:
Ok, and I mostly have them matching the fluid tiers as I used the fluid code as a base to figure out how to add new conduits. However, I made it so that it is a little more in line with the Mekanism pressurized tubes when it comes to how much gas each can transfer by default so that it remains a viable alternative.

PR #5004, opened 2019-01-05
Fix Block Detector

### LINKED ISSUE #5000: Block Detector output is inverted
#### Issue Description:

As a rule, vanilla Minecraft switches emit redstone only when activated. The Block Detector is basically a pressure plate for blocks. So, I would expect a redstone signal to be emitted only when a block is detected.

#### What happens:
A redstone signal is emitted until a block is detected.

A redstone signal to be emitted only when a block is detected.

#### Steps to reproduce:

1. Place Block Detector.
2. Place Redstone near Block Detector.
3. Observe that the redstone is active without a block being detected.

____

- EnderIO: 1.12.2-5.0.38
- EnderCore: 1.12.2-0.5.43
- Minecraft: 1.12.2
- Forge: 1.12.2-14.23.5.2768-universal
- SpongeForge? no
- Optifine? no
- Single player and server.

[latest.log](https://github.com/SleepyTrousers/EnderIO/files/2726002/latest.log)

StackDoubleFlow:
Should fix #5000

### DIFF
--- a/enderio-base/src/main/java/crazypants/enderio/base/block/detector/BlockDetector.java
+++ b/enderio-base/src/main/java/crazypants/enderio/base/block/detector/BlockDetector.java
@@ -135,7 +135,7 @@ public class BlockDetector extends BlockEio<TileEntityPaintedBlock> implements I
 
   @Override
   public int getWeakPower(@Nonnull IBlockState state, @Nonnull IBlockAccess blockAccess, @Nonnull BlockPos pos, @Nonnull EnumFacing side) {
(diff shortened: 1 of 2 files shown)


StackDoubleFlow:
Alright

StackDoubleFlow:
I added it to the config now

HenryLoenwind:
You also need to put in models that have some kind of counter lever that reverses the direction of the upper and lower part of the box that blocks the redstone signal from the internal redstone block.

StackDoubleFlow:
Does the current model not work?

StackDoubleFlow:
Closing since the issue was closed.

PR #5009, opened 2019-01-15
typos, translation clarification

NoX-programer:


NoX-programer:
But I'm speaking Russian and I see:
inaccurate translation: 99, 115 line #
literal, low quality translation: 123, 147, 158, 203 line #
typos: 152, 449 line #
Just annoyed to see this ingame, when tired of fixing lang-file in the mod, I suggest to fix it once and for all.

PR #5036, opened 2019-02-16
Fix owls getting damage when pressing against block above them

Corosauce:
Seems that vanilla collision code doesn't play nice when an entities eye height is the same as the entities full height, this fixes that.

The real fix is the change in the getEyeHeight method, using F3+b you can see the eye height is matched better to the actual owls eyes.

The rest of the code removed is code that attempted to fix this issue which is no longer relevant, removed it within this PR because it felt in scope enough to do so.

Side note: The copied and modified isEntityInsideOpaqueBlock() from an earlier minecraft version and then ported to 1.12 had an additional issue of not updating the use of isOpaqueCube() in it, vanilla 1.12 uses causesSuffocation(), that missing adjustment caused leafs to do damage when they should not, thats all redundant now anyways.


tterrag1098:
Short and sweet. Looks good to me.

PR #5038, opened 2019-02-16
Fix fallen knight not tracking player tightly enough

Corosauce:
First commit fixes a clear bug from porting, shouldContinueExecuting was never executing. This stopped the knight from standing still for 3-4 seconds before continuing to chase you.

Second commit puts the pathing delays more in line with vanilla. Fast enough to track well, but delayed enough to make it not thrash the pathfinder. This resulted in the knight continuing to track you instead of pathing to where you previously were.

I left the path penalty code in shouldExecute because its a good improvement that IMO vanilla zombies should have as well.

Note this doesnt fix the mounted knight spinning in a circle bug, that's a different issue I will do in a different PR.


HenryLoenwind:
> I left the path penalty code in shouldExecute

What do you mean with that? `failedPathFindingPenalty` is not used anywhere...

Issue #5035, opened 2019-02-15
Recipe registry plugin crashed


calloatti:
#### Issue Description:

I get an noticeable delay and an error in the console when using a crafting table or a refined storage crafting grid when doing certain actions involving the use of JEI.

#### What happens:

When trying to close the crafting GUI that pops up when you click on an item in JEI, there is a long delay before the crafting GUI closes.

The delay is longer when using a refined storage crafting grid, and shorter when using a standard crafting table.

When checking the console logs to see if there was anything that could explain the delays, I found an error that seems to blame EnderIO

#### What you expected to happen:

No delay and no error when using JEI in a crafting table/crafting grid

#### Steps to reproduce:

1.Open a crafting table GUI and search for an item in JEI, for example brick:

![2019-02-15_15 49 58](https://user-images.githubusercontent.com/12817862/52877667-6b069080-3139-11e9-9e55-ec1f220c6343.png)

2. Now left click on the brick slabs block to bring up the recipe:

![2019-02-15_15 52 50](https://user-images.githubusercontent.com/12817862/52877774-c3d62900-3139-11e9-943f-39444fd944f2.png)

3. Now **RIGHT CLICK** in the bricks block, to see what recipes use that block. This is the key step that generates the error.

![2019-02-15_15 54 34](https://user-images.githubusercontent.com/12817862/52877873-03047a00-313a-11e9-87b2-37be0d21b4ae.png)

4. Now press the ESC key to exit the JEI GUI, wait a couple of seconds for the dialogbox to close, and check the console log:

```
[15:55:03] [Client thread/ERROR] [jei]: Recipe registry plugin crashed: class crazypants.enderio.base.integration.jei.DarkSteelUpgradeRecipeCategory$1
java.lang.NullPointerException: null
	at crazypants.enderio.base.integration.jei.DarkSteelUpgradeRecipeCategory$1.getWrappers(DarkSteelUpgradeRecipeCategory.java:127) ~[DarkSteelUpgradeRecipeCategory$1.class:?]
	at crazypants.enderio.base.integration.jei.DarkSteelUpgradeRecipeCategory$1.getRecipeWrappers(DarkSteelUpgradeRecipeCategory.java:98) ~[DarkSteelUpgradeRecipeCategory$1.class:?]
[...]

PR #5059, opened 2019-03-02
Fix the NullPointerException in issue #5035

MrMetric:
As far as I can tell, this is a suitable fix. But, I am not familiar with Ender IO's code, so it might not be.


HenryLoenwind:
Yes, that check certainly is needed. Even after the root cause (I added a comment) gets fixed.

PR #5060, opened 2019-03-04
Reduce Memory pressure of UpgradePath

ichttt:
Hi there,
I did some memory analysis on large modpacks, trying to reduce the memory pressure.
I noticed that the HashMap in your DarkSteelUpgradeRecipeCategory is taking up around 150MB of heap space. If you look at the objects inside the map that take, you find that 45MB of the 150MB are just char arrays. This is because the String id can grow quite huge.
My solution is to simply remove the id, and cache the hashcode instead.
The equals check should not be much slower, as it fails fast for different input items

The result is that DarkSteelRecipeManager uses around 45MB less memory.
Some screenshots from eclipse MAT:
Without my changes:
old_dominator
old_retained
With the changes in this PR:
new_retained
new_dominator

### DIFF
--- a/enderio-base/src/main/java/crazypants/enderio/base/handler/darksteel/DarkSteelRecipeManager.java
+++ b/enderio-base/src/main/java/crazypants/enderio/base/handler/darksteel/DarkSteelRecipeManager.java
@@ -20,7 +20,6 @@ import crazypants.enderio.api.upgrades.IDarkSteelItem;
 import crazypants.enderio.api.upgrades.IDarkSteelUpgrade;
 import crazypants.enderio.base.EnderIO;
 import crazypants.enderio.base.lang.Lang;
-import crazypants.enderio.util.StringUtil;
 import net.minecraft.enchantment.Enchantment;
 import net.minecraft.enchantment.EnchantmentHelper;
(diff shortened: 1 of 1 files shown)


enderio-base/src/main/java/crazypants/enderio/base/handler/darksteel/DarkSteelRecipeManager.java
```diff
@@ -188,6 +187,12 @@ public static void addAdvancedTooltipEntries(@Nonnull ItemStack itemstack, Entit
     return result.isEmpty() ? "" : NullHelper.first(result.substring(1), "");
   }
 
+  public static @Nonnull NNList<IDarkSteelUpgrade> getUpgrades(@Nonnull ItemStack stack) {
+    NNList<IDarkSteelUpgrade> list =  UpgradeRegistry.getUpgrades();
+    list.removeIf(upgrade -> !upgrade.hasUpgrade(stack));
```
tterrag1098: I worry that this O(n) code is going to slow down the JEI handler significantly. Have you profiled the difference?

Rongmario:
With the new build out being able to disable dark steel upgrades, this will not be an issue right?

PR #5112, opened 2019-04-23
Fix conduit connector losing priority on unrelated connection changes

codewarrior0:
Sort the tanks list whenever the connection is changed.

### DIFF
--- a/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
+++ b/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
@@ -39,29 +39,12 @@ public class EnderLiquidConduitNetwork extends AbstractConduitNetwork<ILiquidCon
     NetworkTankKey key = new NetworkTankKey(con, conDir);
     NetworkTank tank = new NetworkTank(con, conDir);
 
-    // Check for later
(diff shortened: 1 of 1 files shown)


enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
```diff
@@ -46,9 +46,13 @@ public void connectionChanged(@Nonnull EnderLiquidConduit con, @Nonnull EnumFaci
       sort = true;
     }
 
-    tanks.remove(tank); // remove old tank, NB: =/hash is only calced on location and dir
+    int oldIndex = tanks.indexOf(oldTank);
+    if (oldIndex == -1) {
+      tanks.add(tank);
```
HenryLoenwind: wouldn't it need to sort the list in this case, too?

(I wonder if using a self-sorting list here would not be the better solution.)

codewarrior0: I get the feeling this method is trying to be too clever. Let's just sort the list unconditionally on each change.

HenryLoenwind: Exactly what I was thinking.

PR #5125, opened 2019-05-07
Fixed selection of travel anchor locations when rendered behind the player

### LINKED ISSUE #4773: the ender sword teleports you to an anchor you did not even look at
#### Issue Description:
the ender sword teleports you to an anchor you did not even look at

#### What happens:
the ender sword teleports you to an anchor you did not even look at even if you are not near of an other travel anchor. I also change in the config the behaviour of the anchors to not respond on shift, but this doesn't matter, it happens anyway.

#### Steps to reproduce:

1. have some travel anchors you can use
2. use the travel upgrade on 'the ender'
3. try to teleport without looking to an anchor
...

____

- EnderIO:5.0.26
- EnderCore:1.12.2-0.5.31
- Minecraft: 1.7.12
- Forge:14.23.4.2705

nyronium:
Fixes #4773 

Requires change in EnderCore, which is in this related [pull request #109 for EnderCore](https://github.com/SleepyTrousers/EnderCore/pull/109)

When Optifine or other graphics mods are installed, Tileentities may be rendered even when not in the current frustum. This causes us to be able to teleport to travel anchors that are behind the player, which makes travelling with them really cumbersome.
This PR fixes this by manually culling points by their projected Z coordinate.

Disclaimer:
Unfortunately I could not test this change locally, as some unrelated parts of EnderIO fail to compile on my machine (my java "OpenJDK IcedTea 3.11.0" is more strict about several things like abstract classes etc.). While this is a small change, I could possibly have mistaken +Z and -Z for culling the points, which can be tested easily. If someone can compile EnderIO for me I am happy to test it.

For more information why i propose this change, have a look at the related [issue](https://github.com/SleepyTrousers/EnderIO/issues/4773).

### DIFF
--- a/enderio-base/src/main/java/crazypants/enderio/base/teleport/TravelController.java
+++ b/enderio-base/src/main/java/crazypants/enderio/base/teleport/TravelController.java
@@ -525,8 +525,8 @@ public class TravelController {
         double distance = distanceAndAngle[0];
         double angle = distanceAndAngle[1];
 
-        if (angle < getCandidateHitAngle()) {
-          // Valid hit, sort by distance now.
+        if (distance < closestDistance && angle < getCandidateHitAngle()) {
+          // Valid hit, sorted by distance.
           selectedCoord = bc;
           closestDistance = distance;
         }
--- a/enderio-base/src/main/java/crazypants/enderio/base/teleport/TravelController.java
+++ b/enderio-base/src/main/java/crazypants/enderio/base/teleport/TravelController.java
@@ -545,7 +545,7 @@ public class TravelController {
     if (selectedCoord != null) {
 
       Vector3d blockCenter = new Vector3d(selectedCoord.getX() + 0.5, selectedCoord.getY() + 0.5, selectedCoord.getZ() + 0.5);
-      Vector3d blockCenterScreenSpace = currentView.getScreenPoint(blockCenter);
+      Vector3d blockCenterScreenSpace = currentView.getScreenPoint3D(blockCenter);
       Vector2d blockCenterPixel = new Vector2d(blockCenterScreenSpace.x, blockCenterScreenSpace.y);
 
       Vector2d screenMidPixel = new Vector2d(Minecraft.getMinecraft().displayWidth, Minecraft.getMinecraft().displayHeight);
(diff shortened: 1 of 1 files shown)


HenryLoenwind:
BTW: I'm interested in the compile errors, feel free to ping me on our Discord.

nyronium:
I've adjusted the parameters to match the camera transform matrices. I successfully tested this change locally, backporting is not happening any more.

Btw there is a minor secondary bug, in `getScreenPoint()' (and now also `getScreenPoint3D()`), there is the line `screenPos.scale(1 / transPoint.w);`

When looking straight up, this results in transPoint.w becoming 0 and therefore the return vector will be NaN. Finally, this causes all travel anchors to become valid targets and you are randomly ported to any of those in range.

I could easily prevent this random porting at this point, if desired.

HenryLoenwind:
Yes, looking up should port you to the nearest anchor above you. Would be great if you could fix that one, too.

nyronium:
Okay, I will update here when I had time to test the second fix.

nyronium:
I have investigated the problems with looking straigt up / down.
In the current method there is one main error:
A static up vector (0,1,0) is used to create a view matrix, which becomes a problem if the look vector is parallel to the up vector (missing orthogonal vector), therefore the determinant (and inverse) is not defined. This results in a NaN when calculating inverse matrix to map the points --> strange things happen.

While looking into the issue, I found the current method to be unnecessarily complicated for the task.
Recap: Find anchors near the screen middle to select them.

Currently, the view matrix is stored, inversed frequently and used to map the world points to screen points.

IMO a superior solution is to simply calculate the angle between the given look vector, and the relative vector from eye to the block. This removes the need to store and calculate matrices, eliminates the mapped parameter in the candidate hashmap (would become a hashset), and it inherently differenciates between block before and behind the screen plane.

While testing this I have encountered the block scaling function, where i found that a config parameter ("visualScale") is defaulted to 0.2 instead of a reasonable 1.0 to adjust for a miscalculation in the current scaling function.

I have quickly rewritten the parts I mentioned, which fixes both the backporting problem, the extreme angle problem and theoretically allows for an easy configuration of the scaling parameters. (I have introduced a compensation factor (5) to account for the reduced default config value. It might be favorable to adjust the config default to 1.0 and remove that compensation, but for compatibility I chose the former for this PR.

nyronium:
Side note: The `getScreenPoint3D()` function is not needed for this algorithm, and as there is no EnderIO version using the code, it may be removed from EnderCore.

HenryLoenwind:
Sounds great

nyronium:
Tested and ready to merge from my side.

PR #5144, opened 2019-06-03
Fix totemic capacitors not being able to be enchanted with efficiency and odd levels of efficiency being rounded down

### LINKED ISSUE #5057: Endergy Totemic Capacitor recipe unbalanced... [Can't applied Efficiency Enchant]
#### Issue Description:
I have installed EnderIO and the Endergy Addon, and when I look the recipes and try each new capacitor, the Totemic one (how sincerly is probably one of the most difficult one to make) as just zero benifit over the Crystaline one (how is used into the recipe...). I compare the speed of the machine, the power consumption and the total amount of power capacity, everithing is same... This look totally illogical and unbalanced...

For sure, I can change the recipe my self (And I will probably do for my modpack), but, I think you should look to change somes things about that ;)

#### Steps to reproduce:

1.Install EnderIO + Endergy
2.Test!!
...

____

- EnderIO: 5.0.40
- Endergy: 5.0.40
- EnderCore: 0.5.45
- Minecraft: 1.12.2
- Forge: 14.23.5.2806
- SpongeForge? no
- Optifine? no
- Single Player and/or Server? Single Player, but for sure on server it will be the same!

#### Your most recent log file where the issue was present: I don't think you need logs, because this is not a bug, but possibly an Unexpected Feature!!

Fireztonez

Tfarcenim:
the name of the efficiency enchantment in minecraft is actually "enchantment.digging", not "efficiency", so any attempt to enchant it in survival would fail as that check would never pass.

Closes #5057

### DIFF
--- a/enderio-endergy/src/main/java/crazypants/enderio/endergy/capacitor/ItemTotemicCapacitor.java
+++ b/enderio-endergy/src/main/java/crazypants/enderio/endergy/capacitor/ItemTotemicCapacitor.java
@@ -13,6 +13,7 @@ import crazypants.enderio.base.capacitor.CapacitorHelper;
 import net.minecraft.block.Block;
 import net.minecraft.enchantment.Enchantment;
 import net.minecraft.enchantment.EnumEnchantmentType;
+import net.minecraft.init.Enchantments;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.nbt.NBTTagList;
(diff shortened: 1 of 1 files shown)


Tfarcenim:
image

here's proof

HenryLoenwind:
Weird. Why isn't that a `== net.minecraft.init.Enchantments.EFFICIENCY` check, @EpicSquid ?

Also:
image Seems the original check was meant for the registration name, not the display name.

Tfarcenim:
here's another bug 
image,

EpicSquid:
I huh... I remeber fixing this issue, maybe I just forgot to actually commit it way back. Whatever, it works.

PR #5168, opened 2019-07-08
Multi-tank support for Ender Fluid conduits

### LINKED ISSUE #5005: Fluid conduits not extracting from a machine with multiple outputs
#### Issue Description:

When attempting to extract from a machine with multiple fluid outputs and using a filter, the conduit extracts nothing.

#### What happens:

If the filter is set to pull anything but the first fluid in the machine, it does nothing.

For the fluid to be extracted

#### Steps to reproduce:

1.  Get a machine that can output multiple fluids such as a nuclearcraft electrolyzer. 

2.  Let the machine run until it produces multiple fluid outputs. (eg water to hydrogen, deuterium and oxygen)

3.  Set up a fluid filter to extract any liquid but the 1st (deuterium, oxygen)

4. Watch as it does nothing

____

- EnderIO:  5.0.39
- EnderCore: 0.5.43
- Minecraft: 1.12.2
- Forge:  2796
- SpongeForge? no
- Optifine? no
- Single Player and/or Server? single player

Images:

https://cdn.discordapp.com/attachments/425461908716519425/532350043689058335/2019-01-08_19.07.22.png  https://cdn.discordapp.com/attachments/425461908716519425/532350102027632669/2019-01-08_19.08.53.png https://cdn.discordapp.com/attachments/425461908716519425/532350149423398922/2019-01-08_19.09.05.png

predakanga:
Fixes #5005

### DIFF
--- a/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
+++ b/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
@@ -52,18 +52,24 @@ public class EnderLiquidConduitNetwork extends AbstractConduitNetwork<ILiquidCon
     if (!tank.isValid()) {
       return false;
     }
+
     FluidStack drained = tank.externalTank.getAvailableFluid();
 
     if (!tryExtract(con, conDir, tank, drained)) {
       for (ITankInfoWrapper tankInfoWrapper : tank.externalTank.getTankInfoWrappers()) {
-        drained = tankInfoWrapper.getIFluidTankProperties().getContents();
+        FluidStack toDrain = tankInfoWrapper.getIFluidTankProperties().getContents();
 
-        if (tryExtract(con, conDir, tank, drained)) {
+        // Don't try to drain the same fluid twice
+        if (toDrain != null && toDrain.isFluidEqual(drained)) {
+          continue;
+        }
+
+        if (tryExtract(con, conDir, tank, toDrain)) {
           return true;
         }
       }
     }
-    
+
     return false;
   }
 
--- a/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
+++ b/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
@@ -5,8 +5,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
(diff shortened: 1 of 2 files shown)


enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
```diff
@@ -44,6 +47,11 @@ public void connectionChanged(@Nonnull EnderLiquidConduit con, @Nonnull EnumFaci
     tankMap.remove(key);
     tankMap.put(key, tank);
 
+    hasMultipleTanksSet.remove(key);
+    if (tank.isValid() && !tank.externalTank.getTankInfoWrappers().isEmpty()) {
```
HenryLoenwind: Would `.size() > 1` instead of `!.isEmpty()` make sense here? Do you think assuming that a tank that has a tank list of 1 will stay at 1 and return that one tank's content when asked to "anything"?

enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
```diff
@@ -26,6 +28,7 @@
 
   List<NetworkTank> tanks = new ArrayList<NetworkTank>();
   Map<NetworkTankKey, NetworkTank> tankMap = new HashMap<NetworkTankKey, NetworkTank>();
+  Set<NetworkTankKey> hasMultipleTanksSet = new HashSet<NetworkTankKey>();
```
HenryLoenwind: Could this be a field of `NetworkTank` instead?

(Then the initialization from below would move to NetworkTank's constructor.)

predakanga: Good idea, implemented in 0a5e738

enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/liquid/EnderLiquidConduitNetwork.java
predakanga: It may be a bit paranoid, but the idea here was that `drain` might access a different tank that isn't exposed by `getTankInfoWrappers`.

That is quite unlikely though, so I've changed it to `.size() > 1` in 4f71eaa

HenryLoenwind: yes, I was thinking about that, too. But the chance that a single-tank block correctly implements the properties is much, much higher.

Issue #4238, opened 2017-05-31
[Suggestion] Additional Recipe for ME Conduits


AceologyGaming:
The addition of a recipe to convert ME Cables (AE2) into EIO's ME Conduits would be great.
Numbers-wise it would need to be two ME Cables and three Conduit Binder making two ME Conduits, to match the existing recipe's costs.

Issue #4084, opened 2017-03-09
Dense ME Conduit's recipe is wrong


Xiaminou:
#### Issue Description:
Recipe outputs 8 conduits instead of one, effectively turning 4 ME Conduits into 8 Dense. The recipe should output 1, following the Applied Energistics' Logic.

____
#### Affected Versions (Do *not* use "latest"):

- Minecraft: 1.10.2

NillerMedDild:
That issue has come up again in Minecraft 1.12.2, `EnderIO-1.12.2-5.0.44` - 4 Ender IO ME Conduits can be made into 4 Ender IO Dense Me Conduits. The ratio is 4 Regular to 1 Dense normally.

https://cdn.discordapp.com/attachments/589526583648190465/589569043267190784/unknown.png

HenryLoenwind:
mmmmh... https://github.com/SleepyTrousers/EnderIO/commit/4a4c716555eab367a52793e7ce05cb791018d5fa#diff-38c0adfefbc4985b7d21a31241699932

@EpicSquid any reason you went for different amounts than the original recipes? (see commit from Mar 26, 2017 above)

PR #5174, opened 2019-07-11
Fix an issue with EIO Dense Conduits being better

tyler489:
4 Makes 1....


tyler489:
this will fix #4238 and #4084

Issue #5196, opened 2019-08-07
Apiarist's Shoes Dark Steel Upgrade is uncraftable


YuRaNnNzZZ:
![image](https://user-images.githubusercontent.com/5104603/62654649-ed457400-b968-11e9-8901-61493177e0b8.png)
![image](https://user-images.githubusercontent.com/5104603/62654654-efa7ce00-b968-11e9-9f3d-5e28232839e1.png)

Probably because of the typo in the recipes config (**ee**nderiointegrationforestry)
https://github.com/SleepyTrousers/EnderIO/blob/89b3aefc785c2935cab46bf374aa7d9f367bf7d7/enderio-integration-forestry/src/main/resources/assets/enderio/config/recipes/darksteel_upgrades_forestry.xml#L69

YuRaNnNzZZ:
Same thing with Thaumaturge's Boots upgrade.
![image](https://user-images.githubusercontent.com/5104603/62787882-65797a00-bace-11e9-9470-049b646801cd.png)

Recipe references the "_boots" tag while the actual upgrade ends with "feet"
https://github.com/SleepyTrousers/EnderIO/blob/b4983f13e149851b1122031ab10748fc7fb1973a/enderio-base/src/main/resources/assets/enderio/config/recipes/darksteel_upgrades.xml#L381-L390
![image](https://user-images.githubusercontent.com/5104603/62787916-788c4a00-bace-11e9-83c4-b504d5653f36.png)

tyler489:
Ah... because every other mod calls them boots...

PR #5197, opened 2019-08-07
Spellcheck henry

tyler489:
#5196


HenryLoenwind:
yuo aare copletely richt hear!

PR #5332, opened 2020-02-26
Added a way to automatically minimize/optimize all PNG texture files

BrainStone:
- This PR adds the `optimizePng` task which uses [`optipng`](http://optipng.sourceforge.net/) (which needs to be on the ``PATH``) to optimize all PNG texturefiles.  
  This includes stripping metadata and encoding the image more optimally (like reducing the color pallette and searching for the optimal compression parameters). This compression is completely lossless.
- It also optimizes all existing texture files to show how much it can save.


BrainStone:
If desired I can also add a tiny snippet that will minimize all JSON files in the resulting jars.  
It's a tiny reduction, though it's essentially free. It also should optimize startup times slightly.

PR #5381, opened 2020-05-17
Prevent items ping-pong between I/O inventories

geever:
2 Inverntories on an item conduit-network set to insert&export may ping-pong with items even if the sender has higher input priority. My patch takes source input priority in consideration when building up the sendPriority queue.

### DIFF
--- a/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/item/NetworkedInventory.java
+++ b/enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/item/NetworkedInventory.java
@@ -291,17 +291,29 @@ public class NetworkedInventory {
     if (!canExtract()) {
       return;
     }
+
     List<Target> result = new ArrayList<NetworkedInventory.Target>();
-
     for (NetworkedInventory other : network.getInventories()) {
-      if ((con.isSelfFeedEnabled(conDir) || (other != this)) && other.canInsert()
-          && con.getInputColor(conDir) == other.getCon().getOutputColor(other.getConDir())) {
+      if (!other.canInsert())
+        continue;
+      if (con.getInputColor(conDir) != other.getCon().getOutputColor(other.getConDir()))
+        continue;
 
-        if (ConduitConfig.usePhyscialDistance.get()) {
-          sendPriority.add(new Target(other, distanceTo(other), other.isSticky(), other.getPriority()));
-        } else {
-          result.add(new Target(other, 9999999, other.isSticky(), other.getPriority()));
+      if (other == this) {
+        if (!con.isSelfFeedEnabled(conDir))
+          continue;
+      } else {
+        // If the source have enabled input and have higher or equal priority
+        // than the potential target, don't transfer there
+        if (this.canInsert() && Integer.compare(this.getPriority(), other.getPriority()) >= 0) {
+          continue;
         }
+      }
+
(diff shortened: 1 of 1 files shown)


enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/item/NetworkedInventory.java
```diff
@@ -291,18 +291,30 @@ public void updateInsertOrder() {
...
+      if (con.getInputColor(conDir) != other.getCon().getOutputColor(other.getConDir()))
+        continue;
+
+      if (other == this) {
+        if (!con.isSelfFeedEnabled(conDir))
+          continue;
+      } else {
+        // If the source have enabled input and have higher or equal priority
+        // than the potential target, don't transfer there
+        if (this.canInsert() && Integer.compare(this.getPriority(), other.getPriority()) >= 0) {
```
HenryLoenwind, line 308: Did you check that all conduit changes that have an effect on this trigger a routesChanged()/refreshConnection()?

I think you also need to check if `this`'s extract and insert channels are on the same color.

The bigger issue is that there may be filters in place that prevent looping. You cannot check them here (as there's no item to check), but this would make plenty of filtered networks non-functional.

The correct solution for that would not be trivial. You'd need to do this check when transferring out: If this is not self-feeding but allows insert on the same channel, and the target inventory would extract the item we want to transfer (filter check! Need to check if that inventory would expose the inserted item for extraction (machines don't do that)!) to the same channel and this inventory would accept the item (filter check!), don't transfer. A very expensive check, especially because the filter checks cannot be done without changing the inventory. Which is an issue as many inventories don't support rolling back a change.

And then, why `Integer.compare(this.getPriority(), other.getPriority()) >= 0` when a simple `this.getPriority() > other.getPriority()` would do the same thing?

HenryLoenwind: PS: This needs to support this setup:

1. Chest extract(channel=green) with some gravel inside
2. Chest with insert(channel=green, prio=1, filter=gravel) and extract(channel=green, filter=limited(64 items, gravel))
3. Sagmill with insert(channel=green, prio=0, filter=gravel or flint) and extract(channel=green)
4. Chest with insert(channel=green, prio=-1, filter=flint) and extract(channel=green)
5. Chest with insert(channel=green, prio=-1, filter=sand)

enderio-conduits/src/main/java/crazypants/enderio/conduits/conduit/item/NetworkedInventory.java
```diff
@@ -291,18 +291,30 @@ public void updateInsertOrder() {
...
-          && con.getInputColor(conDir) == other.getCon().getOutputColor(other.getConDir())) {
-
-        if (ConduitConfig.usePhyscialDistance.get()) {
-          sendPriority.add(new Target(other, distanceTo(other), other.isSticky(), other.getPriority()));
-        } else {
-          result.add(new Target(other, 9999999, other.isSticky(), other.getPriority()));
+      if (!other.canInsert())
+        continue;
+      if (con.getInputColor(conDir) != other.getCon().getOutputColor(other.getConDir()))
+        continue;
```
htachoires, line 300: Can you add curly braces ?

PR #5434, opened 2020-09-13
Update en_us.lang

DoctorWeirdGuy:
Forestry Bee filter: Filter Primary/Secondary/Both species button was unlocalized


tyler489:
For Consistency it probably should go here instead

https://github.com/SleepyTrousers/EnderIO/blob/master/enderio-integration-forestry/src/main/resources/assets/enderio/lang/en_us.lang

HenryLoenwind:
Manually applied

PR #5467, opened 2020-10-31
Update items.xml

rhsCZ:
In mc1.12.2 Hopper is not "blockHopper" but only "Hopper". that throw error on startup dw20(2.8.0 pack).


tyler489:
Do you have DietHoppers Installed?

rhsCZ:
No, you mean that is for compatibility reason with this mod?

HenryLoenwind:
blockHopper is the oredict category for hoppers.

DietHoppers replaces the vanilla  hopper and LiteLoader removes the ability to properly replace blocks, that's the usual reason you get an error about this recipe.

PR #5471, opened 2020-11-02
Start Brazilian Portuguese Translation

pauloelr:
I started to translate a few terms to brazilian portuguese, I avoided to translate item names, since they are names. It is partial, I Know but I'm sending it anyway to evaluate acceptance of this PR


pauloelr:
Please let me know if you want me to continue the translation

PR #5535, opened 2021-08-15
Fix Forge maven URL and provide instructions on how to build

embeddedt:
Without the Maven change, builds will fail because `forgeSrc.jar` is unavailable.

I've also added the EnderCore JAR to `.gitignore` so it doesn't show up as an untracked file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ enderio-base/\.checkstyle
 
 **/Corel Auto-Preserve/
 
+## EnderCore lib
+
+/lib/EnderCore*deobf.jar
--- a/README.md
+++ b/README.md
@@ -20,3 +20,11 @@ Below sounds are used under [CC BY 3.0](https://creativecommons.org/licenses/by/
 - https://freesound.org/people/LiamG_SFX/
 - https://freesound.org/people/kuchenanderung1/
 - https://freesound.org/people/170048@virtualwindow.co.za/
+
+### How to compile
+
+1. Clone this repository.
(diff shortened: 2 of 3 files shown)


Pelotrio:
Can you actually build and run the game with this @embeddedt ? I had so many problems trying to do that with the current build gradle.

embeddedt:
It did work at the time, but since then it seems to have broken again, as Player's Maven repository (`http://maven.ic2.player.to`) is now down. Do you know of any other Maven repos with Forestry and/or CoFHCore? Those are the two dependencies erroring for me.

Pelotrio:
I encountered the same problem when I tried building forestry. After searching for a bit I just downloaded it and added it locally. Really annoying how hard it can be these days just to build a mod.

embeddedt:
Yes, and it definitely does not help that the community as a whole is becoming increasingly hostile to any pre-1.16 development.

You guys are doing great work with More Refined Storage. :slightly_smiling_face:

Pelotrio:
Thanks a lot it really is a nice project. Its also nice to see how many times I stumbled over your name just by looking at open prs and comments. I will see if I can build enderio with your changes tomorrow.

embeddedt:
The Forestry Maven repository came back. I tested this again and fixed one other Maven issue. It does build now without problems. `runClient` still has issues because of EnderCore being loaded twice.

This will likely be superseded by #5545 anyways.

Pelotrio:
So you still cant run it? Is there anything that makes it work?

embeddedt:
You should be able to run the reobfuscated JAR produced by `./gradlew build`. It's just userdev (`./gradlew runClient`) that doesn't work properly.

